PR #19, opened 2020-09-10
Add handling of already converted codes

JPercival:


brynrhodes (approved):
Seems like this shouldn't be necessary, but it should be fine for the provider to be defensive like this.

PR #34, opened 2021-03-02
Handle urn prefixes in the BundleRetrieveProvider

### LINKED ISSUE #32: UUIDs in bundles do not resolve when a FHIR Bundle is loaded...
Given an `Observation.subject` reference like this in a Bundle:

```
"subject": {
  "reference": "urn:uuid:0ae954fa-eff8-0d38-c840-3af401abc9fd"
},
```

The evaluator/engine doesn't retrieve the Observation. The FileDataProvider should probably ignore the urn:uuid: scheme just like the bundle retrieve loader does for the patient id.

JPercival:
* Resolves #32 
* Uses new HAPI API for caching FhirContexts
* Fixes some reversed test assert parameters
* Fixes some warnings for static initialization
* Excludes a Java 11+ version of caffeine from auto-updates


bryantaustin13 (approved):
mvn clean install builds and runs tests and passes. 
Looked through code and evaluated changes.
Looks great.

Issue #43, opened 2021-06-10
Remove HAPI dependency from evaluator.measure


csandersdev:
The MeasureEvaluation class in the `evaluator.measure` module does an excellent job of staying model agnostic all the way up to the point where it links some of the parameterized types to HAPI's IBase interface. In order to implement this for a different model (e.g. the IBM FHIR model objects), I would need for that base class to either be parameterized or just left out entirely. It would also be helpful to have the implementations in a separate module from the common classes so that consumers don't pick up a stray HAPI dependency.

JPercival:
Resolved by #44 

Ongoing work with #57 to make this better going forward

PR #44, opened 2021-06-10
#43 - Remove HAPI dependency from evaluator.measure

csandersdev:
I split the implementation classes into a separate module `evaluator.measure.hapi` and parameterized the base type for the model objects.


JPercival:
I created those (mostly) model/version independent classes a while back as an experiment to see what it would take to share all the code across all FHIR versions. In the meantime, we've decided (as in, internally in Alphora) that it's an abstraction too far and not a sustainable pattern across all our projects (cql-engine, evaluator, cqf-tooling, cqf-ruler, etc). There are a lot of subtle issues that arise.

Separately, there's also demand for Measure evaluation in distributed environments such as Spark, a HAPI cluster, etc. and the approach in the current code doesn't play nicely in that space (basically, it assumes all data is in the same database and doesn't have a way to parallelize any of the calculation). So we're looking at doing Measure calculation in two phases:

1. Calculate per-subject Measure reports
2. Aggregate multiple individual reports into a single population-level report.

There's a bit of code written to support that but it's far from complete at this point.

This is all to say that I don't have any issue with your changes or merging them but I expect all of this code to be significantly revamped in the near future and you need to know in case that impacts your long-term strategy. :)

csandersdev:
As you've probably noticed in my other recent PRs, my goal right now is to implement the various clinical reasoning operations inside IBM FHIR. In order to do that, I need implementations that don't use HAPI classes. This seemed like a pretty good start, though I did notice that it was missing the supplemental data code that is present in the cqf-ruler implementation. If you don't feel like it is a good long term base, I could pull it directly into my codebase and work with it from there. I've already got an IBM FHIR implementation on top on the base classes with my small change and it works well enough for what it does. Obviously, I'd love to have the complete implementation that is in CQF ruler when I'm done. I was going to propose making those changes here. Is that something you think you'll do?

JPercival:
The cqf-ruler is derived from the hapi-fhir-jpaserver and for many contexts (Spark, the Atom CQL authoring environment, command-line execution, etc.) that's not workable so in the cql-evaluator we've introduced an interface layer that provides some independence from the underlying datastore. Yes, we're hoping to migrate the implementations of most of the clinical reasoning operations to this repo, including the changes to supplemental data code in Measure evaluate. The code you've proposed changes to was the first attempt at that and it ended up being a dead-end from our perspective.

We didn't have the IBM server specifically in mind, so we are currently still using HAPI object models even though we're not using the jpa layer in the cql-evaluator project. We have tried a few different approaches to provide some degree of version/model independence to maximize the amount of code reuse but have not found a convincing general solution. 

Some things we've tried:

1. Writing all code in terms of R5, and upgrading or downgrading FHIR versions at the input / output. This is the general approach the ig-publisher uses but tends to break down with Resource renames, splits, and merges across versions. Here's an example:

[Resource Parser](https://github.com/HL7/fhir-ig-publisher/blob/e9082fc34e1dadbcd97a8422a75021d589bd1a1c/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/Publisher.java#L4742)

2. Writing all code in terms of fhirPath expressions. This is string-heavy and basically prevents most strong typing in the code. Here's an example of what that looks like:

[Endpoint Converter](https://github.com/DBCG/cql-evaluator/blob/7b58f34905b7ce322c590ec688ee8bae441bf777/evaluator.builder/src/main/java/org/opencds/cqf/cql/evaluator/builder/EndpointConverter.java#L37)

3. Creating a common interface for Resource types, and writing logic against that interface. This is less than ideal because you either get the lowest common denominator for all versions, are a cluttered interface that's the union of all versions with a bunch of "not implemented" bombs waiting to go off. Here's some code that's using that approach:

[Library Adapter](https://github.com/DBCG/cql-evaluator/blob/master/evaluator.fhir/src/main/java/org/opencds/cqf/cql/evaluator/fhir/adapter/LibraryAdapter.java)

4. Generic classes, with function definitions representing the model-specific operations. That's the approach that's in the Measure evaluation code you've updated. This is what I personally think turns out to be the cleanest for most things, but thinking in terms of functions for all the model operations is pretty abstract for your average Java dev:

[R4 Measure Evaluator](https://github.com/DBCG/cql-evaluator/blob/master/evaluator.measure/src/main/java/org/opencds/cqf/cql/evaluator/measure/r4/R4MeasureEvaluation.java#L48)

5. Creating model and version-specific implementations for each operation. Lots of duplicate code and little reuse.

All of that said, we'd absolutely love to be able to support the IBM server (and any others) that wanted to reuse these operations and we'd definitely be open to suggestions on approaches or contributions that would allow that. If you wanted to have a brainstorming session on how to provide model-independent operations I'd be open to it.

csandersdev:
Thanks for the detailed response. You've clearly been wrestling with model independence for a while. 

I get that IBM FHIR isn't the default target for your work and I am certainly not trying to force it into your process. I was just looking for a way to quickly implement the measure evaluation logic without reinventing the wheel. What you have here with the logic class and generics is very functional for that purpose. Without a broader perspective in dealing with multiple FHIR version targets at the same time, I don't know that I have a better suggestion on how to deal with model independence. 

Any type of performance concerns you have are also going to be relevant in my work. We've bandied about the idea of a Spark Cluster to do evaluations, but our initial work opted for an Apache Flink-driven streaming model instead. For us, the Flink evaluators accept a single patient ID and a list of measures to evaluate. Processing all measures for a single patient in the same execution context allows us to cache patient data in the executor that is potentially then reused across multiple measures. We defer to downstream processing to aggregate the reports as needed.

One area that I suspect my project will deviate from the norm is around the subject-list measures. Patient attribution is a complex process in our existing solutions and I'm not sure that relying on the `Patient.generalPractioner` link will be adequate for our use case. If the multiple-patient part of the measure evaluation were separate from the single-patient evaluation logic, would we then be able to keep the single-patient evaluation logic generic and reusable across multiple solutions?

My question about feature parity was mostly trying to get at the long term direction of whether this project becomes the gold-standard for measure evaluation or if it was just prototype/hackathon stuff that wasn't likely to continue. I know you are actively evolving other parts of the project, but the measure evaluation part was unclear to me given feature differences with cqf-ruler. It sounds like you do plan to use this project. If you wanted, I could take a stab at trying to bring the evaluation logic in line with cqf-ruler with regard to supplemental data. I don't want to iterate on code that is a dead end though.

Back on the topic of model independence, it is clear you are thinking about the bigger picture of what it would mean to be independent in all areas of the cql-evaluator. If that were possible, it would be very cool. An earlier part of my project involved creating command line tools to execute CQL and evaluate measures along with resource adapters for ZIP files, filesystem bundles, and live FHIR server resources. All of that overlaps with what you've done here in cql-evaluator and I like a lot of what you did better. It isn't a high priority for me to replace that work though.

JPercival:
> Processing all measures for a single patient in the same execution context allows us to cache patient data in the executor that is potentially then reused across multiple measures. We defer to downstream processing to aggregate the reports as needed.

This how the Spark implementations I've done work as well.

>  If the multiple-patient part of the measure evaluation were separate from the single-patient evaluation logic, would we then be able to keep the single-patient evaluation logic generic and reusable across multiple solutions?

It's at least easier to do so, yes. There are at least two considerations there. One is that the single-patient logic is significantly simpler if the aggregation is done separately. The other is that, as you point out, determining the appropriate list of subjects for evaluation varies from platform-to-platform and use-case to use-case. The logic in the cqf-ruler currently assumes that every patient on the local server (for a given practitioner) should be evaluated. We really need a pluggable interface to get that list instead. This is a long-standing issue:

[How Is "patient-list" Being Determined?](https://github.com/DBCG/cqf-ruler/issues/40)

> My question about feature parity was mostly trying to get at the long term direction of whether this project becomes the gold-standard for measure evaluation or if it was just prototype/hackathon stuff that wasn't likely to continue.

Yes, this is intended to be the gold-standard eventually. I actually hadn't intended to publish the measure module yet since it was prototype/hackathon stuff, but I see that I did. That was my mistake.

> I could take a stab at trying to bring the evaluation logic in line with cqf-ruler with regard to supplemental data. 

That would be appreciated! That said I dug a little deeper into the IBM java representation of the FHIR types and compared them to the HAPI types, with an eye towards seeing if we could plausibly support both models. I tried one more idea which was to create templates for the logic that used different imports. For example:

template:
```
import  ${package}.Measure

public String getId(Measure m) {
   return m.getId();
}
```

IBMOperation.java
```
import  com.ibm.fhir.model.resource.Measure

public String getId(Measure m) {
   return m.getId();
}
```

HAPIOperation.java
```
import  org.hl7.fhir.r4.model.Measure

public String getId(Measure m) {
   return m.getId();
}
```

There are definitely some things I like better about the IBM representation such as the integrated Builders. Unfortunately, there are enough differences between the two models that the templating approach won't work in general.

The only paths forward I can see that allow us to be model-independent are to:

1. Generate our own interfaces / models for the FHIR structure definitions, implement the operations in terms of those interfaces, and provide mappings to the IBM/HAPI/other models. Given that there are already several Java representations to choose from and given that this project is primarily funded by contracts intended to further reference implementation work, it doesn't seem like a good idea for us to generate our own. That overlaps significantly with work that's already done and creates potential issues with integrating with the rest of the reference implementation ecosystem. It's also something that we, frankly, don't have the bandwidth to support at this point.

2. Take a functional approach like I did with the Measure evaluation. This effectively allows us to bite off the model and version independence in operation-sized chunks. We _had_ decided against this, but we were mainly thinking in the context of FHIR versions rather than completely different models. However, potentially getting a broader base of support for these implementations (specifically, integration in the IBM) is enough to get us to reconsider. We're talking about it internally and experimenting with some of the other operations we've defined but have not come to a conclusion yet. There are some significant downsides in that it's not a typical paradigm for Java.

So, in the meantime, I'm going to accept this PR and I'd accept any PRs updating the Measure logic to what's currently there in the ruler, and I'm going to iterate on this Measure code a bit myself for some project deliverables I have, with the assumption that this will be the gold-standard going forward.

csandersdev:
The IBM model is definitely a different beast. Most notably, the model objects are immutable and must be converted to/from builder objects to change properties. The Builders also default to validate on build() which can be troublesome when you are used to creating partial objects and here are also some nuances around dealing with FHIR types that overlap with Java types (e.g. String, Boolean) that require care and attention.

I think it is fair to say that if it ends up costing more code to be model agnostic than it would take to simply write model-specific implementations of the logic, then we've probably made the wrong choice. If this ends up being a burden, let me know and I'll go a different direction.

I'll see what I can do with the supplemental data implementation. I spent some time in the cqf-ruler code and I think I have a handle on what it is doing. I'm curious, though, is the expected behavior described in an IG somewhere? I poked around the IGs that I'm familiar with and I didn't see anything that talked about the relationship between Measure.supplementalData codes and Patient.extension values that is being established by the SDE logic. There are some things, such as setting the cqf-measureInfo.measure = "http://hl7.org/fhir/us/cqfmeasures/" + report.getMeasure() that feel solution-specific to me. Is the SDE collection something that we should maybe delegate to a different provider?

With regard to this current PR, the build failure is something related to Javadoc that has been eluding me. I'll spend some more time on it and see if I can get that figured out.

PR #46, opened 2021-06-16
Expression evaluation module

JPercival:
* Partial implementation of the cpg $cql operation, which specific emphasis on FhirPath support (thanks @jreyno77 !)
* Tests for same


brynrhodes (approved):
Looks great, builds locally, tests pass

jreyno77 (approved):

PR #47, opened 2021-06-18
Fix codeutils enumeration support

JPercival:
* Add support for getting codes from Enumeration
* Fixed accidentally removed CodeUtilTests
* Update to HAPI 5.4.1


sliver007 (approved):
Looks good. Builds locally and test passes.

PR #69, opened 2021-11-03
Fix aggregate report

mdnazmulkarim:
fix contained merge for subject-list type measure reports


bryantaustin13 (approved):
LGTM. Builds clean and tests pass.

evaluator.measure-hapi/src/main/java/org/opencds/cqf/cql/evaluator/measure/r4/R4MeasureReportAggregator.java
```diff
@@ -68,22 +71,46 @@ protected void mergeContained(MeasureReport carry, MeasureReport current) {
...
         }
+        carry.getContained().clear();
 
         for (Resource resource : current.getContained()) {
-            if (resource.hasId()) {
-                if (!resourceIds.contains(resource.getId())) {
-                    resourceIds.add(resource.getId());
-                    carry.getContained().add(resource);
+            if (resource.hasId() && resource.getResourceType().equals(ResourceType.List)) {
+                if (resourceMap.containsKey(resource.getId())) {
```
JPercival, line 85: The lists here may have different Ids. You'll have to pick the correct list based on the population list reference in the population group.

JPercival (changes requested):

JPercival (approved):
Needs some tweaks on subject-list merging, but I'll work on that.

Issue #70, opened 2021-11-06
Support cql-options in the evaluator


brynrhodes:

JPercival:
Resolved by #71

PR #71, opened 2021-11-06
#70: Added support for translator options in the file- and FHIR-based…

brynrhodes:
… library content providers


evaluator.cql2elm/src/main/java/org/opencds/cqf/cql/evaluator/cql2elm/content/LibraryContentProvider.java
```diff
@@ -32,4 +36,19 @@
     default InputStream getLibrarySource(VersionedIdentifier libraryIdentifier) {
         return getLibraryContent(libraryIdentifier, LibraryContentType.CQL);
     }
+
+    /**
+     * Provides translator options for an evaluation context.
+     * @return CqlTranslatorOptions
+     */
+    default CqlTranslatorOptions getTranslatorOptions(VersionedIdentifier vi) {
```
JPercival, line 44: Is the plan for the cql-translator as well? The LibraryContentProvider interface was meant to augment whatever the translator was doing until the translator got direct support for loading ELM.

evaluator.builder/src/main/java/org/opencds/cqf/cql/evaluator/builder/library/CqlFileLibraryContentProviderFactory.java
```diff
@@ -96,7 +96,7 @@ public LibraryContentProvider create(String url, List<String> headers) {
         }
 
         if (resourceDirectory.getAbsoluteFile().isDirectory()) {
-            return FileUtils.listFiles(resourceDirectory, new String[] { "cql" }, true);
+            return FileUtils.listFiles(resourceDirectory, new String[] { "cql", "json" }, true);
```
JPercival, line 99: This is no longer needed given that the translator options are part of the cli options, correct?

brynrhodes: Yes, missed this revert. Fixed

PR #73, opened 2021-12-10
Removing JAXB dependencies from the engine.

vitorpamplona:
- Removed JAXB dependencies from the engine.
- Corrected an evaluator (ProperlyIncludes) class name to match the generated class. 
- Adjusted tests to avoid XML loads.


vitorpamplona:
Folks, since we have already merged the Engine's removal of the JAXB, we need this PR approved in order to work with the $apply/PlanDefinition one.

brynrhodes (dismissed):
Seeing an error in the build that it cannot construct a LibraryMapper. Seems to be related to the change in the names from Properly to Proper?

JPercival:
Don’t merge this one yet. I need to get 1.3.1 out the door.

JPercival (approved):

PR #93, opened 2022-04-20
Fix Unresolved ValueSet Expansion

jreyno77:
Added check for unexpanded valuesets with expansion logic and throw error

Added tests for change

### DIFF
--- a/evaluator.fhir/src/test/resources/org/opencds/cqf/cql/evaluator/engine/util/r4/TestBundleValueSetsTwo.json
+++ b/evaluator.fhir/src/test/resources/org/opencds/cqf/cql/evaluator/engine/util/r4/TestBundleValueSetsTwo.json
@@ -1,0 +1,39 @@
+{
+    "resourceType": "Bundle",
+    "id": "test-bundle-two-valuesets",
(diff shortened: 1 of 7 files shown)


evaluator.fhir/src/test/resources/org/opencds/cqf/cql/evaluator/engine/util/r4/TestBundleValueSetsTwo.json
```diff
@@ -0,0 +1,39 @@
+{
+    "resourceType": "Bundle",
+    "id": "test-bundle-two-valuesets",
```
JPercival, line 3: Let's give this an id / filename that indicates that this content is actually for testing logic expansion.

JPercival (changes requested):

Jinip (dismissed):
Add a check for an expansion parameter named "naive" with a value "true", which are value sets expanded by the EnsureExecutableValueSet tooling function, not a terminology server. Return an info level error message that says something to the effect of "codes expanded without a terminology server, some results may not be correct"

JPercival (approved):

PR #103, opened 2022-06-07
PlanDefinition Test infrastructure overhaul and Care Plan with Questionnaires 

vitorpamplona:
This is a lot, but bear with me

**The Problem**

1. MockFhirDal had several hardcoded loaders, making it difficult to add new tests. 
2. Those hardcoded loaders were making certain tests pass, even though they should be failing. 
3. A Convoluted test class was also not helpful.
4. Test resources were all mixed. 
5. Tests were using ShallowEquals, which didn't catch many small inconsistencies on the generated Care Plan. 
6. PlanDefinitions that led to Questionnaires were not supported. 
7. PlanDefinitions that use non Canonical references were not supported. 
8. Goals were not being mapped on $apply
9. $apply was creating new IDs (random UUIDs), instead of delegating it to the DAL

**The Solution**

Complete overhaul of the test infrastructure: 
1. New MockFhirDal that stores loaded objects in memory. 
2. New Base test class to reset the PlanDefinition Processor at every test. 
3. New folder structure for test cases. 
4. Tests now compare exported JSONs and must match completely. 

Tests were fixed: 
1. UUID ids are not expected anymore. 
2. Identifiers were corrected to the appropriate references. 
3. Duplicated JSONs were deleted. 

Questionnaire Definitions were introduced. 
1. Test cases were based on [Android FHIR #1403](https://github.com/google/android-fhir/issues/1403). 

URI-based tasks were introduced. 
1. This is just a basic implementation until we have more test cases


JPercival (approved):
Awesome. Love me a good overhaul.

PR #126, opened 2022-07-23
Add concept text to contained SDE Observation to support backwards co…

rob-reynolds:
…mpatibility

Add display of same to same to be consistent with system and code


c-schuler (approved):
Looks good. Builds locally. Tests passed

PR #129, opened 2022-08-01
Update sde for non-retrieve or in-cql resource return

mdnazmulkarim:
The cql-created resources will be included in contained.
Related [issue](https://github.com/DBCG/cqf-ruler/issues/574)


rob-reynolds (approved):
Tested locally, produces expected results.

Tests pass.

LGTM

PR #133, opened 2022-08-09
support instance resources in referred library or in another expressi…

mdnazmulkarim:
We are inspecting Instance element in cql. The cons is it will add execution overhead.
```
define "SDEResourceConstructionTest":
  SDE."Resource Construction Test"
  
define "SDEResourceConstructionTestInlib":
  "Resource Construction Test Inlib"  
  
define "Resource Construction Test Inlib":
  FHIR.Observation {
    id: FHIR.id { id: 'observation-sdetest-' + ToString(Now()) },
    status: FHIR.ObservationStatus { value: 'final' },
    code: FHIR.CodeableConcept { coding: { FHIR.Coding { code: FHIR.code { value: 'days-since-appointment' } } } },
    subject: FHIR.Reference { reference: FHIR.string { value: Patient.id.value }},
    value: FHIR.integer { value: Abs(difference in days between end of SDE."Last Appointment".period and ToDateTime(Today())) }
  }  
```
SDE:
```
define "Resource Construction Test":
  FHIR.Observation {
    // id: FHIR.id { id: 'observation-sdetest-' + ReplaceMatches(ToString(Now()), ':', '') },
    id: FHIR.id { id: 'observation-sdetest-' + ToString(Now()) },
    status: FHIR.ObservationStatus { value: 'final' },
    code: FHIR.CodeableConcept { coding: { FHIR.Coding { code: FHIR.code { value: 'days-since-appointment' } } } },
    subject: FHIR.Reference { reference: FHIR.string { value: Patient.id.value }},
    value: FHIR.integer { value: Abs(difference in days between end of "Last Appointment".period and ToDateTime(Today())) }
  }
```


evaluator.measure/src/main/java/org/opencds/cqf/cql/evaluator/measure/common/MeasureEvaluator.java
```diff
@@ -437,8 +439,21 @@ protected void evaluateSdes(List<SdeDef> sdes) {
 
     // consider more complex expression in future
     private void inspectInstanceEvaluation(SdeDef sdeDef, ExpressionDef expressionDef) {
```
JPercival, line 441: I don't understand why we need to do inspection of the resources or of the ExpressionType. Why is `getEvaluatedResources` not enough to distinguish between constructed and retrieved resources?

rob-reynolds (approved):
Tested locally.

Tests passed.

LGTM

PR #134, opened 2022-08-11
Fix instance resource expression mr

mdnazmulkarim:


rob-reynolds (approved):
Passes local tests.

LGTM

PR #142, opened 2022-08-27
update sde inspect instance resource logic

mdnazmulkarim:
Still not decided whether we are doing this or not!


JPercival:
Superseded by #152

PR #143, opened 2022-08-28
Removing heavy resources from the main fhir lib

vitorpamplona:
These are ~50MB each of xml files.

### DIFF
--- a/evaluator.fhir/pom.xml
+++ b/evaluator.fhir/pom.xml
@@ -48,23 +48,11 @@
         </dependency>
         <dependency>
             <groupId>ca.uhn.hapi.fhir</groupId>
-            <artifactId>hapi-fhir-validation-resources-dstu3</artifactId>
-        </dependency>
-        <dependency>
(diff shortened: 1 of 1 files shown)


JPercival:
IIRC those are loaded at runtime whenever you're using the FhirPath engine, which we were using in a couple different places within the evaluator:

```
FhirContext context = FhirContext.forR4();
DefaultProfileValidationSupport validationSupport = new DefaultProfileValidationSupport(context);
HapiWorkerContext workerContext = new HapiWorkerContext(context, validationSupport);
FHIRPathEngine fhirPathEngine = new FHIRPathEngine(workerContext);
```

The cql-engine also supports FhirPath, so one possibility is to just use the cql-engine for those same use cases. I'll have to go back and see if we still need this anywhere because it's been a couple years.

JPercival:
Seems we killed of the FhirPath code in favor of using the cql-engine already. So I think we're good to merge!

PR #147, opened 2022-09-08
Update the evaluator to use the 2.1.0-SNAPSHOT engine

brynrhodes:


pom.xml
```diff
@@ -310,7 +310,7 @@
                 <groupId>com.github.ben-manes.caffeine</groupId>
                 <artifactId>caffeine</artifactId>
                 <version>${caffeine.version}</version>
-                <scope>test</scope>
+                <!--<scope>test</scope>-->
```
JPercival, line 313: This needs to be changed at the `cli` project level.

pom.xml
```diff
@@ -443,7 +443,7 @@
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-source-plugin</artifactId>
-                    <version>3.2.1</version>
+                    <version>3.1.0</version>
```
JPercival, line 446: Why was this downgraded?

brynrhodes: Because I'm grasping at straws trying to figure out why a simple dependency won't resolve.

Issue #177, opened 2022-10-25
Support NPM library loading in the evaluator


brynrhodes:
Need to support the ability to load libraries from NPM packages in the evaluator

brynrhodes:
Included in 2.3.0

PR #178, opened 2022-10-25
#177: Added NPM library loading support to the evaluator

brynrhodes:


JPercival (commented):
Is there a version of this that doesn't include 200+ image, json, javascript, etc. files? Almost none of the test resources have much to do with namespaces AFAICT.

PR #188, opened 2022-10-27
Fix dstu3 apply dynamicValue resolution

barhodes:


c-schuler (approved):
Looks good. Builds and tests pass locally

PR #192, opened 2022-11-02
Fix resource type resolution from contained resources

maimoonak:
The fhir-path specs allows to use [contained resources into plan-definition](https://hl7.org/fhir/R4/plandefinition-example.json.html) for better organization of dependent resources like ActivityDefinition. This is especially helpful and clean when using multiple ActivityDefinitions for a PlanDefinition. The current code despite allowing and parsing the local-ids (at L355) in method applyActivityDefinition does not work with local contained resources and fails before due to not being able to parse resource name


vitorpamplona:
This makes sense to me. But can we add a test case to make sure we don't break this behaviour afterwards?

maimoonak:
@vitorpamplona Added tests as well. Kindly check.

JPercival (approved):

PR #193, opened 2022-11-03
Fixes for downstream CDSHooks update 

c-schuler:
Fixes to enable resolution for https://github.com/DBCG/cqf-ruler/issues/539
Fixes issue with LibraryProcessor not accounting for reference-styled patient IDs (e.g. Patient/example-patient)
Fixes issue with PlanDefinitionProcessor not resolving expression evaluation correctly - assumes the expression will be an identifier
Also cleaned up a bunch of warnings


barhodes (approved):
Some of that cleanup should likely go into the dstu3 and r5 PlanDefinitionProcessors, where applicable.  Other than that it looks good to me.

barhodes (approved):

PR #243, opened 2023-02-06
NULL contextValue in BundleRetrieveProvider:retrieve()

dkayiwa:
This pull request is for helping me find out why `contextValue` is always `NULL` in the `BundleRetrieveProvider:retrieve()` method callback.


dkayiwa:
@JPercival @barhodes am trying to use the CQL engine via this evaluator library in the ANC DAK project for OpenMRS. But the `contextValue` parameter in `BundleRetrieveProvider:retrieve()` is always null. Could it be that i have not correctly set up the required data? I have set up this pull request for you to reproduce this problem and help point me to where am doing the right thing. This is the [plandefinition](https://github.com/WorldHealthOrganization/smart-anc/blob/master/input/resources/plandefinition/plandefinition-ANCDT17.json), [libraries](https://github.com/WorldHealthOrganization/smart-anc/tree/master/input/resources/library), and [valuesets](https://github.com/WorldHealthOrganization/smart-anc/tree/master/input/vocabulary/valueset) that i have bundled in `library-bundle.json`.

Thanks in advance, for spending your precious time on this!

gracepotma:
@brynrhodes friendly reminder ping ;) We are completely blocked without your help to move forward here.

JPercival:
CQL has the concept of [context](https://cql.hl7.org/02-authorsguide.html#context) that provides an implicit filter for `Retrieve`s.

As an example, here's a `Retreive` of some `Encounter` resources in `Patient` context:

```cql
context Patient

define "Encounters":
   [Encounter]
```

That translates to a FHIR Query that looks like `Encounter?subject=Patient/XYZ` where `XYZ` is the patient currently in context. The set of FHIR Resources that support Patient context roughly corresponds to those in the [Patient compartment](https://build.fhir.org/compartmentdefinition-patient.html)

You must set the appropriate context value when evaluating with the cql-engine, that's not automatically inferred. If no context value is supplied, the `contextValue` will be null.

Certain operations like `$evaluate-measure` and `$apply` have an input parameter that corresponds to the context and context value, typically `subject`. If you run `$evaluate-measure` with a `subject=Patient/123`, the context value for the `Patient` context will be `123`.

Some resources don't have a relationship to the `Patient` resource. For example, in FHIR, Medications are not Patient-specific and are outside the Patient compartment:

```cql
context Patient

define "Medications":
   [Medication]
```
That will return all Medications. In these cases the content value is always `null` because it is not relevant.

If you're not using one of the pre-built operations you need to set the context value directly on the cql-engine yourself:

```cql
var context = new org.opencds.cqf.cql.engine.execution.Context(library);
context.setContextValue("Patient", "123");
```

dkayiwa:
Thanks @JPercival for the detailed response!

In this pull request, i am dealing with the `Observation` resource, which is in the [Patient compartment](https://build.fhir.org/compartmentdefinition-patient.html), and therefore supports the Patient context.

I am also using the `$apply` operation on a PlanDefinition, to generate a patient CarePlan. Since i supply a patient, i would expect it to be reflected in the context value. But it is null.

Do you have some time to look at the contents of this pull request and point me at where i could possibly not be doing the correct thing?

gracepotma:
@JPercival or @brynrhodes any further thoughts on Daniel's clarifying questions above?

evaluator.plandefinition/src/test/resources/org/opencds/cqf/cql/evaluator/plandefinition/r4/anc-dak/data-bundle.json
```diff
@@ -0,0 +1,236 @@
...
+			},
+			"request": {
+				"method": "PUT",
+				"url": "Encounter/helloworld-patient-1-encounter-1"
+			}
+		},
+		{
+		     "resource": {
+		        "resourceType": "Patient",
+		        "id": "Patient/5946f880-b197-400b-9caa-a3c661d23041",
```
barhodes, line 221: This id is likely the cause of your issues.  Remove the "Patient/" from the id and it should be able to find this patient to use as the context.

dkayiwa: I have just done and still get a context value of null. In case this is helpful, below are the values of the various parameters in `BundleRetrieveProvider:retrieve`:

context: `Encounter`
contextPath: `encounter`
contextValue: `null`
dataType: `Observation`
templateId: `http://hl7.org/fhir/StructureDefinition/Observation`
codePath: `code`
codes: `null`
valueSet: `http://fhir.org/guides/who/anc-cds/ValueSet/anc-b8-de17`
datePath: `null`
dateLowPath: `null`
dateHighPath: `null`
dateRange: `null`

brynrhodes: Apologies for the delays here, looking at this briefly, I notice that the "context" is "Encounter", is there a reason you are choosing an Encounter context rather than a Patient context?

dkayiwa: I do not call this method directly. It is the CQL engine/framework that calls it: https://github.com/cqframework/clinical-reasoning/blob/v2.6.0/evaluator.fhir/src/main/java/org/opencds/cqf/cql/evaluator/engine/retrieve/BundleRetrieveProvider.java#L50-L55

brynrhodes: Right, but that's already after the engine is running with the parameters that it has been set up with. This would be earlier in the process where you are actually setting up the engine context.

brynrhodes: Looking at the test I don't see anything that would set it up for Encounter context. We have an engineer looking at this now and will get back to you as soon as possible. Again, apologies for the delays here.

dkayiwa: Thank you so much for looking at it. I am anxiously looking forward to the response! 😊

brynrhodes: @dkayiwa , I'm able to successfully run the ANCDT17 test. Is that the expected result, or is the test failing for you?

dkayiwa: @brynrhodes  that is another thing that i wanted bring up after this has been sorted. The test passes regardless of the patient observation values. In other words, the output care plan is always exactly the same whether the CQL evaluations return true or false.

brynrhodes: I see, so there's something more fundamental going on. We'll keep digging.

dkayiwa: Thank you so much @brynrhodes 👍

brynrhodes: @dkayiwa , we think we found and addressed an issue with this, in that the PlanDefinition $apply processor wasn't creating actions if the PlanDefinition.action didn't have a definition or any child actions. That fix has been merged to master (https://github.com/cqframework/clinical-reasoning/pull/266) However, I'm not sure we ever reproduced the null context though, so I'd like to see if you can recreate the behavior with the latest from master if that's alright?

brynrhodes: We pulled the tests from this PR into that one, so you should be able to just check and see whether you're getting expected behavior.

PR #248, opened 2023-02-17
Feature cqf setup eval

mdnazmulkarim:
WIP:
to do:

- > Repository is FhirContext agnostic
- >Engine Context needs an empty constructor
- >Parameters : should we interface 'org.opencds.cqf.cql.evaluator.fhir.util.Parameters' or
     'org.hl7.fhir.instance.model.api.IBaseParameters' to maintain FhirContext agnostically.
- > Do we need DataProviderFactory interface enhancement to accomodate 'create(repository)'


evaluator.fhir/src/main/java/org/opencds/cqf/cql/evaluator/fhir/util/FhirResourceLoader.java
```diff
@@ -22,9 +22,11 @@ public FhirResourceLoader(FhirContext context, Class<?> clazz, List<String> dire
         List<String> locations = new ArrayList<>();
 
         directoryList.forEach(dir -> {
+            System.out.println("dir:" + dir);
```
JPercival, line 25: Let's remove the System.out.println statements.

PR #265, opened 2023-03-08
Check method overload

mdnazmulkarim:


JPercival (approved):

evaluator/src/main/java/org/opencds/cqf/cql/evaluator/engine/execution/TranslatingLibraryLoader.java
```diff
@@ -151,4 +170,21 @@ protected Library translate(VersionedIdentifier libraryIdentifier) {
     }
   }
 
+  private boolean hasOverloadedFunctions(Library library) {
+    Set<String> functionNames = new HashSet<>();
+    if (library != null && library.getStatements() != null) {
+      for (ExpressionDef ed : library.getStatements().getDef()) {
+        if (ed instanceof FunctionDef) {
+          FunctionDef fd = (FunctionDef) ed;
```
JPercival, line 178: The overloads only clash if they have the same number of parameters:

```
DoThing(value : Integer) == DoThing(value : String)

DoThing(value : Integer) != DoThing(value : String, value2 : Whatever)
```

JPercival: class FunctionSig() {
  private final String name:
  private final int numArguments;
  
  public FunctionSig(String name, int numArguments) {
     this.name = name;
     this.numArguments = numArguments;
  }
  
  @Overide
  public boolean equals(Object other) {
     if (other == null) {
        return false;
     }
     
     FunctionSig func = (FunctionSig)other;
     
     if (func == null) {
        return null;
     }
     
     return this.name.equals(other.name) && this.numArguments == other.numArguments;
  }
  
  @Override
  public int hashCode() {
    int start = 17
    return start + name.hashCode() * 31 + numArguments * 31;
  }
}

JPercival: Set<FunctionSig> functionNames = new HashSet<>();

var sig = new FunctionSig(fd.getName(), fd.arguments.size());
funtionNames.contains(sig);

JPercival (approved):

PR #273, opened 2023-03-22
Resolve issues for hapi-fhir integration

barhodes:
Refactor and cleanup to resolve issues integrating operations into hapi-fhir

### DIFF
--- a/cqf-fhir-utility/src/main/java/org/opencds/cqf/fhir/utility/Searches.java
+++ b/cqf-fhir-utility/src/main/java/org/opencds/cqf/fhir/utility/Searches.java
@@ -7,6 +7,8 @@ import java.util.Map;
 
 import ca.uhn.fhir.model.api.IQueryParameterType;
 import ca.uhn.fhir.rest.param.StringParam;
+import ca.uhn.fhir.rest.param.TokenParam;
+import ca.uhn.fhir.rest.param.UriParam;
 
 public class Searches {
 
@@ -19,7 +21,7 @@ public class Searches {
   }
 
   public static Map<String, List<IQueryParameterType>> byUrl(String url) {
-    return builder().withToken("url", url).build();
+    return builder().withUri("url", url).build();
   }
 
   public static Map<String, List<IQueryParameterType>> byName(String name) {
@@ -38,6 +40,15 @@ public class Searches {
       return this.values;
     }
 
+    SearchBuilder withString(String name, String value) {
+      if (values == null) {
+        values = new HashMap<String, List<IQueryParameterType>>();
+      }
+      values.put(name, Collections.singletonList(new StringParam(value)));
+
+      return this;
+    }
+
     SearchBuilder withToken(String name, String value) {
       // var paramValue = new StringParam(value);
       // var paramList = new ArrayList<IQueryParameterType>();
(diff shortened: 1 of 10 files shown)


cqf-fhir-utility/src/main/java/org/opencds/cqf/fhir/utility/Searches.java
```diff
@@ -38,14 +40,32 @@ public Map<String, List<IQueryParameterType>> build() {
       return this.values;
     }
 
+    SearchBuilder withString(String name, String value) {
```
chalmarm, line 43: Can the method name be more descriptive or comments on the method would help.

evaluator.activitydefinition/src/main/java/org/opencds/cqf/cql/evaluator/activitydefinition/r4/ActivityDefinitionProcessor.java
```diff
@@ -33,10 +39,42 @@ public ActivityDefinitionProcessor(Repository repository) {
...
+          theResourceType.getSimpleName(), theUrl));
+    }
+
+    return (R) searchResult.getEntryFirstRep().getResource();
+  }
+
+  @Override
+  public ActivityDefinition resolveActivityDefinition(IIdType theId, String theCanonical,
+      IBaseResource theActivityDefinition) throws FHIRException {
+    var baseActivityDefinition = theActivityDefinition != null ? theActivityDefinition : null;
```
chalmarm, line 58: IS the line redundant, the line below is anyways we are checking for null.

evaluator.activitydefinition/src/main/java/org/opencds/cqf/cql/evaluator/activitydefinition/r5/ActivityDefinitionProcessor.java
```diff
@@ -35,10 +41,42 @@ public ActivityDefinitionProcessor(Repository repository) {
...
+    if (!searchResult.hasEntry()) {
+      throw new FHIRException(String.format("No resource of type %s found for url: %s",
+          theResourceType.getSimpleName(), theUrl));
+    }
+
+    return (R) searchResult.getEntryFirstRep().getResource();
+  }
+
+  @Override
+  public ActivityDefinition resolveActivityDefinition(IIdType theId, String theCanonical,
```
chalmarm, line 58: Question: I have seen this method resolveActivityDefinition in 3 different places. Are all the implementations the same, if so can they be abstracted to any super implementation. For me them looked exactly same unless I am not seeing something.

Even the method above searchRepositoryByUrl. 

Please correct me or let me know if we need to discuss.

barhodes: The resource types that are needed here are version specific.  Hence the reason for the abstract method only in the Base class and the very similar implementation in each version.

barhodes: The searchRepositoryByUrl function could potentially go into a version specific helper class and I briefly thought about doing that but decided to wait until we have a better place to put such things, i.e. once we have everything shifted from the evaluator packages into the cqf-fhir packages.

JPercival (approved):

PR #275, opened 2023-03-27
ANCDT17 Test Case

c-schuler:
- Updated test case to latest 
- Now including params from apply operation request into action parameter resolution


barhodes (approved):
The failing test is due to a bug in one of the test resources that I have fixed in a separate PR.

PR #298, opened 2023-05-02
Bug caching overloads

JPercival:
* Default to using ELM with at least "Overloads" level of signature generation
* Remove "CacheAwareModelManager" - the upstream ModelManager code now supports caching directly.
* Fixed relevant tests


evaluator/src/main/java/org/opencds/cqf/cql/evaluator/cql2elm/model/CacheAwareModelManager.java
barhodes, line 1: This class is referenced in hapi-fhir in the BaseClinicalReasoningConfig.  We'll need to inc the version for this.

barhodes (approved):

PR #306, opened 2023-05-15
Async security context

Capt-Mac:
In order to fix security context dropping when in threaded mode, added delegation of security context to executor


evaluator.measure-hapi/pom.xml
```diff
@@ -64,6 +64,10 @@
       <version>3.0.0-PRE3-SNAPSHOT</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
JPercival, line 67: Should no longer need this dependency.

JPercival (approved):

PR #312, opened 2023-06-13
FHIR Repository API (in-memory)

c-schuler:
Added an in-memory repository with search functionality and tests.

Question for reviewer:
Is the search functionality needed for the Bundle repository? (It is an easy add if it is needed)


barhodes (approved):

c-schuler:
@barhodes I appreciate the contributions to this PR. However, I would like to know why these changes are necessary. It would have been nice to have you review the code and request changes that I could apply. One change I added that you rolled back was ignoring the provided resource ID for the create operation (as the [spec outlines](http://hl7.org/fhir/R4/http.html#create)). What is the reasoning to not ignore the ID? How does create differ from the update operation?

barhodes:
> @barhodes I appreciate the contributions to this PR. However, I would like to know why these changes are necessary. It would have been nice to have you review the code and request changes that I could apply. One change I added that you rolled back was ignoring the provided resource ID for the create operation (as the [spec outlines](http://hl7.org/fhir/R4/http.html#create)). What is the reasoning to not ignore the ID? How does create differ from the update operation?

I ran into issues using the IdDt type when matching that were resolved by using the Ids helper like we did in the TestRepository.  Looks like I got a little overzealous in merging those on the create and missed the random id getting created.  I'll correct that.  The rest of the changes were all done to ensure compatibility with hapi-fhir so the search parameters work correctly when using the Clinical Reasoning module against Smile CDR.

c-schuler:
Thanks for the info! So this PR is kind of in a weird state as you are both a contributor and a reviewer. Should I review your code or will someone else be providing that review?

barhodes:
Yeah, I've asked JP to review.  You are more than welcome to review as well.

evaluator.fhir/src/main/java/org/opencds/cqf/cql/evaluator/fhir/test/TestRepository.java
```diff
@@ -0,0 +1,27 @@
...
+
+import java.util.List;
+
+import org.hl7.fhir.instance.model.api.IBaseBundle;
+import org.opencds.cqf.cql.evaluator.fhir.repository.InMemoryFhirRepository;
+
+import ca.uhn.fhir.context.FhirContext;
+
+/**
+ * This class has been deprecated. Use InMemoryFhirRepository instead.
```
dkayiwa, line 11: @barhodes could there be a reason why you did not use the [deprecated annotation](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Deprecated.html) for this class?

barhodes: Only my ignorance of that annotation.  I will get it added in.  Thanks!

PR #343, opened 2023-09-15
Fix profile for $apply response

barhodes:
Remove CPG Strategy profile from RequestGroups.
Only add to instantiatesCanonical the direct PlanDefinition that is being instantiated by a RequestGroup (not any nested PlanDefinitions)

### DIFF
--- a/cqf-fhir-cr/src/test/resources/org/opencds/cqf/fhir/cr/plandefinition/dstu3/child-routine-visit/child_routine_visit_careplan.json
+++ b/cqf-fhir-cr/src/test/resources/org/opencds/cqf/fhir/cr/plandefinition/dstu3/child-routine-visit/child_routine_visit_careplan.json
@@ -5,11 +5,6 @@
     {
       "resourceType": "RequestGroup",
       "id": "ChildRoutineVisit-PlanDefinition-1.0.0",
-      "meta": {
-        "profile": [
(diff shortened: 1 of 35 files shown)


cqf-fhir-cr/src/test/resources/org/opencds/cqf/fhir/cr/plandefinition/dstu3/child-routine-visit/child_routine_visit_careplan.json
```diff
@@ -5,11 +5,6 @@
     {
       "resourceType": "RequestGroup",
       "id": "ChildRoutineVisit-PlanDefinition-1.0.0",
-      "meta": {
```
JPercival, line 8: Question: What was the impetus for removing the `profile`? Just the fact that declaring your own profile is an anti-pattern in FHIR?

JPercival (approved):

cqf-fhir-cr/src/test/resources/org/opencds/cqf/fhir/cr/plandefinition/dstu3/child-routine-visit/child_routine_visit_careplan.json
barhodes: Given that  a resource can adhere to any number of profiles, it was decided that the calling source should be responsible for adding in and validating expected profiles.

PR #348, opened 2023-09-20
Add unit test code coverage requirements

JPercival:
* Add jacoco to the maven build
* Integrate code cov with github actions


JPercival:
Did some testing with this on #380. Seems to work!

PR #353, opened 2023-10-02
repository search by id

mdnazmulkarim:
Search by id visits the intended execution path, tested and screenshoot from debug:
image


cqf-fhir-cr/src/test/java/org/opencds/cqf/fhir/cr/measure/r4/InMemoryRepositoryTest.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    @Test
+    public void testRead() {
+        IBaseResource res = repository.read(Library.class, new IdType("Library/LibraryEvaluationTest"), null);
+        assertEquals(res.getIdElement().getIdPart(), "LibraryEvaluationTest");
+    }
+
+    @Test
+    public void testSearchWithId() {
```
JPercival, line 39: Issue: This test doesn't validate that the fast path is used. Searching by id already worked. This code is supposed to make that faster.

JPercival: Issue: This test is also in the wrong place. It should be under the `utility` folder, the same as the repository implementation. This is a test of repository behavior, not a test of Measure behavior.

cqf-fhir-cr/src/test/resources/org/opencds/cqf/fhir/cr/measure/r4/bundle/test-bundle.json
```diff
@@ -0,0 +1,207 @@
+{
```
JPercival, line 1: Issue: This content is much bigger than it needs to be. We only need the following resource:

```
{
   "resourceType": "Library",
   "id": "LibraryEvaluationTest"
} 
```

In fact, there's no reason to create a JSON resource at all. It's so simple it could just be part of the Java code.

JPercival (changes requested):

cqf-fhir-utility/src/main/java/org/opencds/cqf/fhir/utility/repository/InMemoryFhirRepository.java
```diff
@@ -125,6 +125,14 @@ public <B extends IBaseBundle, T extends IBaseResource> B search(
                 .values();
         if (searchParameters == null || searchParameters.isEmpty()) {
             resourceList.forEach(builder::addCollectionEntry);
+        } else if (searchParameters.containsKey("id")) {
+            for (IBaseResource resource : resourceList) {
```
JPercival, line 129: Issue: This implementation is incorrect. The point is that the resourceMap is _already_ indexed by Id. You don't need to scan all the resources. You can simply look at the one you need from the index.

JPercival (changes requested):

cqf-fhir-cr/src/test/java/org/opencds/cqf/fhir/cr/measure/r4/InMemoryRepositoryTest.java
```diff
@@ -37,9 +38,14 @@ public void testRead() {
 
     @Test
     public void testSearchWithId() {
-        Map<String, List<IQueryParameterType>> map = new HashMap<>();
-        map.put("id", Collections.singletonList(new ReferenceParam("Library/Example1")));
-        IBaseBundle bundle = repository.search(IBaseBundle.class, Library.class, map, null);
+        Map<String, List<IQueryParameterType>> mapLibrary = new HashMap<>();
+        mapLibrary.put("id", Collections.singletonList(new ReferenceParam("Library/Example1")));
```
JPercival, line 42: Issue: the SearchParameter is `_id`, not `id`. The tests should reflect that.

cqf-fhir-utility/src/main/java/org/opencds/cqf/fhir/utility/repository/InMemoryFhirRepository.java
JPercival: Issue: The SearchParameter is `_id`, not `id`. The implementation and the tests should reflect that.

JPercival (approved):

Issue #344, opened 2023-09-18
R4MeasureReportBuilder bug on buildSDE if reportType not set for population


Capt-Mac:
Description: 
$evaluate-measure parameter reportType is not defaulting to the correct value of "summary" when the subject parameter is not populated. When running against a population and reportType is not explicitly mentioned as "population" it will throw an illegal exception. 

Steps:
1. run evaluate measure operation against >1 patients
2. omit the ReportType parameter
3. R4MeasureReportBuilder.buildSDE throws illegal argument error saying evaluating results >1 and reportType=individual

Expected outcome:
Processor logic would automatically set reportType to "population" if subject parameter is omitted. Per [spec](https://build.fhir.org/operation-measure-evaluate-measure.html)

Version being used CR: 3.0.0-PRE7-SNAPSHOT

PR #354, opened 2023-10-03
Measure service subject reporttype bugs

Capt-Mac:
- Update r4MeasureService from hapi-fhir MeasureService
- Fix [bug](https://github.com/cqframework/clinical-reasoning/issues/344) where reportType=null does not cause error
- Add capability for r4 evaluateMeasure to handle practitioner & practitioner group via R4SubjectProvider
- Fix [bug ](https://github.com/alphora/dqm-platform/issues/295) where summary subject references were not being populated by Group/Practitioner references 
- update Measure Given tests to use MeasureService instead of processor

### DIFF
--- a/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4MeasureProcessor.java
+++ b/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4MeasureProcessor.java
@@ -96,9 +96,7 @@ public class R4MeasureProcessor {
         var subjects =
                 subjectProvider.getSubjects(actualRepo, evalType, subjectIds).collect(Collectors.toList());
 
-        if(StringUtils.isNotBlank(reportType)){
-            evalType = MeasureEvalType.fromCode(reportType).get();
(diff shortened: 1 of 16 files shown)


cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4MeasureProcessor.java
```diff
@@ -97,6 +97,12 @@ protected MeasureReport evaluateMeasure(
 
         var subjects =
                 subjectProvider.getSubjects(actualRepo, evalType, subjectIds).collect(Collectors.toList());
+        /*
```
barhodes, line 100: This comment block is no longer needed, yes?

barhodes (approved):

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4MeasureProcessor.java
Capt-Mac: removed

Issue #358, opened 2023-10-10
Population Measure crashes when subjectIds is null or empty


vitorpamplona:
if `subjectIds` is null or empty, it crashes the app. 

Keep in mind that `orElse` evaluates even if its object is filled. `subjectIds.get(0) == null` should probably be the last expression in this if. 

https://github.com/cqframework/clinical-reasoning/blob/0fe8e3afd23437d0afe09677ebfe22d36723e066/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4MeasureProcessor.java#L88

Same happens in R5

PR #359, opened 2023-10-11
Fixes crash on measure evaluate when subjectIds is null

vitorpamplona:
Fixes https://github.com/cqframework/clinical-reasoning/issues/358


JPercival (approved):

JPercival:
Huh. Well that's a silly one.

There's a conflict with `master`. Could you resolve that and apply formatting `./mvnw spotless:apply` and I'll merge?

vitorpamplona:
hum.. i think there is a bug in this spotless apply action. Its not actually running

JPercival:
Permission issue on the github action.

PR #362, opened 2023-10-12
Fixes Measure Evaluate Start and End times

vitorpamplona:
Something like this: 

```kotlin
    val measureReport =
      fhirOperator.evaluateMeasure(
        measureUrl = "http://fhir.org/guides/who/anc-cds/Measure/ANCIND01",
        start = "2020-01-01",
        end = "2020-01-31",
        reportType = MeasureEvalType.SUBJECT.toCode(),
        subjectId = "charity-otala-1",
        practitioner = "jane",
      )
```

went from outputting 

```json
  "period": {
    "start": "2020-01-01T00:00:00+00:00",
    "end": "2020-01-31T23:59:59+00:00"
  },
```

to 

```json
  "period": {
    "start": "2020-01-01T00:00:00+00:00",
    "end": "2020-01-31T00:00:00+00:00"
  },
```

after the upgrade to 3.0. 

After lots of debug, I believe the following lines set the precision to seconds without considering if the last day is closed or open, which causes the subsequent operations to be based on the 0 hour and the output to match that.


vitorpamplona:
@JPercival looks like you changed these lines recently. You might be aware of this potential issue.

JPercival:
We have a couple DateTime bugs in-flight, one was resolved with https://github.com/cqframework/clinical_quality_language/pull/1259.

I'm going to take another look at this once https://github.com/cqframework/clinical_quality_language/pull/1269 lands.

JPercival:
@Capt-Mac - Can you take a look at this? DateTime fixes were merged into CQL.

Based on the spec here I'd say that the millisecond-level output is correct:
https://build.fhir.org/operation-measure-evaluate-measure.html

Capt-Mac:
@vitorpamplona  I took a look at your changes to R4MeasureReportBuilder.java

R4MeasureProcessor.buildMeasurementPeriod does correctly default interval to expected value, which is what the CQL will use to actually calculate results:

- test submitted (2014-01-01, 2014-12-31)
- observed results (2014-01-01T00:00:00, 2014-12-31T23:59:59)

This matches spec definition https://build.fhir.org/operation-measure-evaluate-measure.html

"The end of the measurement period. The period will end at the end of the period implied by the supplied timestamp. E.g. a value of 2014 would set the period end to be 2014-12-31T23:59:59 inclusive"

MeasureReport.Period however has different requirements https://www.hl7.org/fhir/R4/datatypes.html#Period
and R4MeasureReportBuilder.getPeriod does correctly assume period based on provided values to $evaluate-measure.

Though for clarity, I do think these values should match

whatever value is passed in gets put on https://www.hl7.org/fhir/R4/datatypes.html#dateTime

example: passed in => returned
yyyy-mm-dd => yyyy-mm-ddT00:00:00, 
yyyy-mm-ddT23:59:59 => yyy-mm-ddT23:59:59

JPercival:
Superseded by deciding to handle timezones via the FHIR timezone client header.

Issue #334, opened 2023-08-30
encounter parameter is not passed during the CQL engine execution


dkayiwa:
For [all these ANC measures](https://github.com/WorldHealthOrganization/smart-anc/tree/master/input/resources/measure), the encounter parameter does not get passed to the engine execution. Therefore, when evaluating `Last(Split(encounter, '/'))` as seen [on this line](https://github.com/WorldHealthOrganization/smart-anc/blob/master/input/resources/measure/measure-ANCIND01.json#L71), using the `org.opencds.cqf.cql.engine.elm.execution.ParameterRefEvaluator` which has a name of `encounter` in the `org.opencds.cqf.cql.engine.elm.execution.SplitEvaluator.internalEvaluate()`, the `getStringToSplit().evaluate(context)` [method call](https://github.com/cqframework/cql-engine/blob/master/engine/src/main/java/org/opencds/cqf/cql/engine/elm/execution/SplitEvaluator.java#L46) returns NULL.

This is the data bundle that i am using: https://github.com/openmrs/openmrs-module-cql/blob/main/api/src/test/resources/org/openmrs/module/cql/measure/Observation-ANCIND7-Bundle.json

This is the measure: https://github.com/WorldHealthOrganization/smart-anc/blob/master/input/resources/measure/measure-ANCIND07.json

This is the library: https://github.com/WorldHealthOrganization/smart-anc/blob/master/input/resources/library/library-ANCIND07.json

And the test looks like this: https://github.com/openmrs/openmrs-module-cql/blob/main/api/src/test/java/org/openmrs/module/cql/AncDakTest.java#L131-L154

PR #363, opened 2023-10-13
Implemented parameter support for measure evaluation

c-schuler:
Fix for #334 

@dkayiwa please verify this resolves the issue you're seeing. I added a test that uses the data/artifacts you referenced in the issue.


cqf-fhir-cr/src/test/resources/org/opencds/cqf/fhir/cr/measure/r4/ANC/resources/library/library-ANCBaseConcepts.json
```diff
@@ -0,0 +1,53 @@
...
+  }, {
+    "type": "depends-on",
+    "display": "Value set Iron supplement",
+    "resource": "http://fhir.org/guides/who/anc-cds/ValueSet/iron-supplement"
+  }, {
+    "type": "depends-on",
+    "display": "Value set Folic acid supplement",
+    "resource": "http://fhir.org/guides/who/anc-cds/ValueSet/folic-acid-supplement"
+  } ],
+  "content": [ {
```
JPercival, line 43: We're trying to reduce bloat in the repo by stripping away unneeded content/metadata for libraries. Check out the example here:
https://github.com/vitorpamplona/clinical-reasoning/blob/master/cqf-fhir-cr/src/test/resources/org/opencds/cqf/fhir/cr/measure/r4/FHIR347/resources/library/FHIR347.json

Notice that the content is now a reference to the CQL content. This means we don't have to do the base64 encode/decode while we're editing.

JPercival (changes requested):

cqf-fhir-cr/src/test/resources/org/opencds/cqf/fhir/cr/measure/r4/ANC/resources/library/library-ANCBaseConcepts.json
JPercival: There's a cqf-tooling operation to do that automatically here:
https://github.com/cqframework/cqf-tooling/pull/450

JPercival: I went ahead and did this.

JPercival (approved):

Capt-Mac (approved):

PR #403, opened 2024-01-12
Measurescorer group bug fix

### LINKED ISSUE #399: MeasureScorer null pointer exception
When a user has a multi rate measure that attempts to build a measureScorer per group, if one group.id does not match it tries to match on group.population.id, which is not universally populated and receives a "null pointer" exception

here is an example report that fails:
[measurereport-multirate-nopopid.json](https://github.com/cqframework/clinical-reasoning/files/13906675/measurereport-multirate-nopopid.json)

Expected behavior:
measurereport scorer calculates measurescore without error

Capt-Mac:
closes #399


cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4MeasureReportScorer.java
```diff
@@ -41,14 +41,18 @@ protected MeasureScoring getGroupMeasureScoring(
                     break;
                 }
             }
-            // Match by group's population id
-            if (mrgc.getPopulation().size() == entry.getKey().populations().size()) {
+            // Match by group's population id, if group-ids are not present
+            if ((mrgc.getPopulation().size() == entry.getKey().populations().size())
```
JPercival, line 45: thought: We ought to add validation for the Measure on the frontend of evaluation such that group or population is required. If both are absent we have no way to line up the scores.

JPercival (approved):

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4MeasureReportScorer.java
Capt-Mac: It currently errors out if missing both group & population ids, the test "noids" in measureScorerTest captures this invalid scenario. It could probably be a more meaningful error than the current "no measure scoring value set"

PR #404, opened 2024-01-17
Enable ErrorProne and CheckStyle

### LINKED ISSUE #405: Enable ErrorProne for the build
[ErrorProne](https://errorprone.info/) is a static analysis library from Google that catches many common programming errors. It's part of the CQL build upstream when enabled with the proper profile. We should adopt it on this project as well to catch similar issues.

JPercival:
* Enable ErrorProne
  * Fixes for all errors
  * Warnings disabled for now, will revisit in the future
* Enable CheckStyle for Google naming conventions
  * Palantir format already enforced by spotless
  * We were starting to mix HAPI conventions into the codebase, sometimes within the same class and even using different conventions for different method parameters
  * Fixes for all cases using HAPI conventions, resetting to Google conventions
  
  Resolves #405


barhodes (approved):
Looks like there are still issues in the benchmark package.  Changes all look good to me, though.

PR #409, opened 2024-01-30
Enforce ids for elements which are required in the CQF Measure Profile

JPercival:
* Validates Ids are required from most elements of Measure
* Updates existing test content to be compliant with the above
* Adds tests for invalid cases
* Validation logic reordered in both DSTU3 and R4 so that they both fail with same exceptions for the same type of content errors.
* The profile we're checking against: https://build.fhir.org/ig/HL7/cqf-measures/StructureDefinition-measure-cqfm.html


barhodes (approved):

JPercival:
Resolved #412

PR #410, opened 2024-01-31
Measurescorer bug fix

### LINKED ISSUE #400: MeasureScorer does not calculate population exclusions or exceptions correctly
Measure Scorer currently only scores the values within a group based on Numerator/Denominator, but does not take into account the removal of exceptions and exclusions. This means that MeasureScores that should be removing counts due to an exception or/and exclusion are not doing so and scores are inaccurately recorded in the MeasureReport

Observed behavior:
Measure Report with population scores of:
ip:2, den:2, den-except:0, den-exclu:1, num: 1 => 0.5

Expected behavior:
ip:2, den:2, den-except:0, den-exclu:1, num: 1 => 1.0
den - den-exception = total den

Capt-Mac:
closes #400

adding extensions to each group and/or stratum that calculates measureScore by accumulating values that match designed set logic for respective populations in a measureReport resource. 

https://build.fhir.org/ig/HL7/cqf-measures/measure-conformance.html#proportion-measure-scoring

Any existing logic that aggregates or scores based off of only numerator/denominator will need to be updated in order to calculate the measure-score accurately, especially where denominator exception cases are used.


barhodes (changes requested):
All these .DS_Store files should not be added to the repo.

barhodes (approved):

cqf-fhir-cr/src/test/java/org/opencds/cqf/fhir/cr/measure/r4/MeasureScorerTest.java
```diff
@@ -84,6 +155,15 @@ MeasureReportGroupComponent group(MeasureReport measureReport, String id) {
...
+    public MeasureReport.MeasureReportGroupStratifierComponent getStratumById(
+            MeasureReport measureReport, String groupId, String stratifierId) {
+        var group = group(measureReport, groupId);
+        return group.getStratifier().stream()
+                .filter(g -> g.getId().equals(stratifierId))
+                .findFirst()
+                .get();
+    }
+
     public List<Measure> getMyMeasures() {
```
JPercival, line 167: question: does this work for Measures that have some other basis than "Patient"? Have we validated it works for encounters, as an example? How about cohort Measures?

JPercival (approved):

cqf-fhir-cr/src/test/java/org/opencds/cqf/fhir/cr/measure/r4/MeasureScorerTest.java
```diff
@@ -73,8 +90,62 @@ void testScore_error_noids() {
...
-        var e = assertThrows(IllegalArgumentException.class, () -> scorer.score(measureScoringDef, measureReport));
-        assertEquals("No MeasureScoring value set", e.getMessage());
+        scorer.score(measureScoringDef, measureReport);
+        // when denominator =0, no score should be added to report
+        assertNull(group(measureReport, "DataCompleteness").getMeasureScore().getValue());
+    }
+
+    @Test
+    void testScore_noExtension() {
+        var measureUrl = "http://content.alphora.com/fhir/uv/mips-qm-content-r4/Measure/multirate-noext";
```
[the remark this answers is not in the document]
Capt-Mac: only reason this should not render is if it is not able to score a denominator, in which case it wouldn't score it. I don't think an error is needed

cqf-fhir-cr/src/test/java/org/opencds/cqf/fhir/cr/measure/r4/MeasureScorerTest.java
Capt-Mac: cohort measures should not apply because they don't calculate a measure score (no num or den), it does however add extensions still to the measureScore.group for cohorts. This should be removed

Capt-Mac: encounter measure like CMS111 has a test case and does not currently add the new extensions and does not calculate measureScore

Capt-Mac: CMS111 is continuous variable, so like cohort, doesn’t apply to measureScoring

DischargedonAntithromboticTherapyFHIR is proportion using that same extension for encounter and does appropriately get scored using newly added logic.

PR #413, opened 2024-02-02
Fixes for malformed codes and valuesets

JPercival:
* Fixes null pointer exceptions when codes are incompletely specified
* Fixes null pointer exceptions when valuesets are incompletely specified
* The above is accomplished by ignoring (i.e. filtering out) partially specified codes


JPercival:
resolves #414

rob-reynolds (approved):

PR #427, opened 2024-02-16
Fixing CLI Tests due to case sensitivity

JPercival:


JPercival:
Closed in favor of #428

PR #441, opened 2024-03-19
Fix issue with item generation for hidden items

barhodes:
With this change, when generating Questionnaire items from a Profile, if a differential element does not have a type, the type from the snapshot (if available) will be used. Alternatively, if the element has child elements defined the type is ignored and a group item is created for the children elements.


rob-reynolds (approved):
Tested.  Expected results.

Capt-Mac (approved):

PR #449, opened 2024-04-15
Add input parameter resolution to $populate operation

barhodes:
Add the InputParameterResolver to the PopulateRequest which resolves the input parameters to resources to allow them to be referenced within an expression, e.g. %subject.id

### DIFF
--- a/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/questionnaire/populate/PopulateRequest.java
+++ b/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/questionnaire/populate/PopulateRequest.java
@@ -1,5 +1,7 @@
 package org.opencds.cqf.fhir.cr.questionnaire.populate;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 import ca.uhn.fhir.context.FhirVersionEnum;
 import org.hl7.fhir.instance.model.api.IBaseBundle;
 import org.hl7.fhir.instance.model.api.IBaseOperationOutcome;
@@ -36,6 +38,7 @@ public class PopulateRequest implements IQuestionnaireRequest {
             Boolean useServerData,
             LibraryEngine libraryEngine,
             ModelResolver modelResolver) {
+        checkNotNull(libraryEngine, "expected non-null value for libraryEngine");
         this.operationName = operationName;
         this.questionnaire = questionnaire;
         this.subjectId = subjectId;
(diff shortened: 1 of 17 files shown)


cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/questionnaire/populate/PopulateRequest.java
```diff
@@ -23,38 +23,39 @@ public class PopulateRequest implements IQuestionnaireRequest {
...
         this.subjectId = subjectId;
         this.parameters = parameters;
         this.bundle = bundle;
+        this.useServerData = useServerData;
         this.libraryEngine = libraryEngine;
         this.modelResolver = modelResolver;
         this.fhirVersion = questionnaire.getStructureFhirVersionEnum();
         this.defaultLibraryUrl = resolveDefaultLibraryUrl();
+        inputParameterResolver = libraryEngine == null
+                ? null
```
JPercival, line 50: question: As far as I can tell, the libraryEngine is a required parameter. The implementation of `getParameters` depends on it existing. Rather than a `null` check, is it possible to use a `notNull` assertion?

JPercival (approved):

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/questionnaire/populate/PopulateRequest.java
barhodes: Something like this?

JPercival: Yes, but was hoping to get both code paths unified. IOW, what's the use case where we supply a `inputParametersResolver` independent of the `libraryEngine`. Can we always just create a `inputParametersResolve` off of Library engine? Or vice-versa?

JPercival: Granted, this is a nitpick, so feel free to say no. Just wondering if we can unify it all.

barhodes: In the PlanDefinition $apply case I allow for a custom resolver to be passed in, but we don't really have a use case for that yet.  Definitely doable but I would want to spend a little more time on it so think it would be best in another PR.

PR #453, opened 2024-05-02
multi-measure service capability

### LINKED ISSUE #454: Feature Request: Multi-Measure Processing for MeasureService
$evaluate-measure operation currently fulfills the requirement of evaluating a single measure for a chosen subject population, however when a user requires the same subject population to be evaluated against several measures, it requires subsequent requests and redundant data layer calls.

This is especially useful for any users that are performing quality reporting for a specific population over several measures and can now handle in a single request.

I have also logged this with a standards change request: 
https://jira.hl7.org/browse/FHIR-45420

Capt-Mac:
closes #454 

evaluate measure service to intake several measures and output a bundle of measureReports per Measure.


barhodes (approved):

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4MultiMeasureService.java
```diff
@@ -0,0 +1,147 @@
...
+
+        var subjectProvider = new R4RepositorySubjectProvider();
+
+        var processor = new R4MeasureProcessor(repo, this.measureEvaluationOptions, subjectProvider);
+
+        R4MeasureServiceUtils r4MeasureServiceUtils = new R4MeasureServiceUtils(repository);
+        try {
+            r4MeasureServiceUtils.ensureSupplementalDataElementSearchParameter();
+        } catch (NotImplementedOperationException e) {
+            log.warn(
```
JPercival, line 73: issue: I don't think the exception here can be hit. There's a catch inside the util class. I'd suggest just removing this exception block.

JPercival (approved):

PR #467, opened 2024-05-29
 NOJIRA-public_expand_value_set

ibaisi:
make expandValueSet public

### DIFF
--- a/cqf-fhir-utility/src/main/java/org/opencds/cqf/fhir/utility/visitor/r4/KnowledgeArtifactPackageVisitor.java
+++ b/cqf-fhir-utility/src/main/java/org/opencds/cqf/fhir/utility/visitor/r4/KnowledgeArtifactPackageVisitor.java
@@ -291,7 +291,7 @@ public class KnowledgeArtifactPackageVisitor {
         });
     }
 
-    public ValueSet expandValueSet(
+    public void expandValueSet(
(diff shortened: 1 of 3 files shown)


cqf-fhir-utility/src/main/java/org/opencds/cqf/fhir/utility/visitor/r4/KnowledgeArtifactPackageVisitor.java
```diff
@@ -352,6 +352,8 @@ protected void expandValueSet(
                         "Terminology Server expansion failed for: " + valueSet.getId(), ex.getMessage());
             }
         }
+
+        return valueSet;
```
TahaAttari, line 356: Question: If we're going to change it from `void` > `ValueSet`, would it be better to have it expand a copy of the input `ValueSet` and return the copy? 

If the method returns something then I feel like it would cause confusion if the method mutated the input _and_ returned a output at the same time?

ibaisi: I reverted the return type as it would require to make ValueSet clonal which is not.

barhodes (changes requested):
Let's keep parity between versions.  Make sure this change is made in the Dstu3 and R5 versions of this class.

barhodes (approved):

PR #468, opened 2024-05-30
[APHL-954] recursive action check

TahaAttari:


barhodes (approved):
Seems like we should be looking at action.definition to ensure we get all possible dependencies.  Are those being handled elsewhere?  Test coverage in these classes is a bit low.  Would be grand to add a few more tests.

TahaAttari:
Yes, good catch 🙏🏾  we should be checking for action.definitionCanonical 👍🏾 I'll make the update

PR #475, opened 2024-06-08
Add support for the SDC Observation Extract Category extension

### LINKED ISSUE #474: SDC Observation Extract Category extension is not supported by the $extract operation

barhodes:
When extracting using observation based extraction, if the item has an SDC Observation Extract Category extension, use the value of that extension for the category of the extracted Observation.  If no extension is present the category will default to "survey".

Closes #474


JPercival (changes requested):
Can we get the test coverage up to 80% on changed code?

JPercival (approved):

PR #483, opened 2024-06-26
run $expand on type instead of instance if url is preset

TahaAttari:


barhodes (changes requested):
This does not seem correct to me.  The expansionParameters object being passed in here is a contained Parameters resource in the rootSpecificationLibrary.  How would that singular Parameters object contain the url parameter for each ValueSet we are expanding?

barhodes (approved):

PR #491, opened 2024-07-22
multi-measure feature enhancements

Capt-Mac:
new feature expansion to multi-measure evaluate request:
- individual measureReport mode, that allows for individual MeasureReport creation for each subject identified via the subject parameter
- consolidated practitioner parameter capability into subject parameter, for a single population parameter. This allows for simplified explanation and use of api
- canonicalUrl Measure parameter that allows for both measureId and measureUrl to be used to identify the Measure to use.
- stubbed in identifier search logic for measures once implemented

closes https://github.com/cqframework/clinical-reasoning/issues/492

### DIFF
--- a/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/utils/R4MeasureServiceUtils.java
+++ b/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/utils/R4MeasureServiceUtils.java
@@ -185,6 +185,7 @@ public class R4MeasureServiceUtils {
 
     public Measure resolveByIdentifier(String identifier) {
         List<IQueryParameterType> params = new ArrayList<>();
+        Map<String, List<IQueryParameterType>> searchParams = new HashMap<>();
         Bundle bundle;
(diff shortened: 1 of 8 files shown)


Capt-Mac:
need testing data for proxy repository testing

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/utils/R4MeasureServiceUtils.java
```diff
@@ -150,4 +159,99 @@ public Reference getReporter(String reporter) {
...
+                throw new IllegalArgumentException(msg);
+            }
+            return (Measure) bundle.getEntryFirstRep().getResource();
+        } else {
+            var msg = String.format("Measure Identifier: %s, found no matching measure resources", identifier);
+            throw new IllegalArgumentException(msg);
+        }
+    }*/
+
+    public List<Measure> getMeasures(
```
barhodes, line 218: Take a look at the common.ResourceResolver.  It may be useful here.

barhodes (approved):

Issue #494, opened 2024-07-23
Support measure content release


brynrhodes:
Support the creation of a measure content release.

The $release operation needs to be able to be used to package a Measure Library (i.e. a Library whose components are Measures, and not owned).

PR #495, opened 2024-07-23
#494: Added measure support for the release operation

brynrhodes:
This is a draft PR that implements Measure support in the Release operation. Additional changes include:

* Improved dependency tracing in the Release operation to include non-owned components
* Added support for non-owned components
* Added a title element to the KnowledgeArtifactAdapter interface
* Added a descriptor function to KnowledgeArtifactAdapter to provide a "display" element when writing a relatedArtifact

Current gaps in this implementation:

* MeasureAdapter is only provided for R4, need DSTU3 and R5 support
* The test is hardcoded to the ecqm-content-qicore-2024-subset repository, need to isolate the test resources and provide assertions
* The "descriptor" implementation doesn't work properly right now
* The relatedArtifact for CodeSystem resources that cannot be resolved should use the expansionParameters in the manifest, if it is present
* CodeSystem and ValueSet references should also result in a "system-version" or "canonical-version" entry in the expansion parameters, respectively

### DIFF
--- a/cqf-fhir-utility/src/main/java/org/opencds/cqf/fhir/utility/Canonicals.java
+++ b/cqf-fhir-utility/src/main/java/org/opencds/cqf/fhir/utility/Canonicals.java
@@ -134,7 +134,7 @@ public class Canonicals {
     public static String getUrl(String canonical) {
         checkNotNull(canonical);
 
-        if (!canonical.contains("/")) {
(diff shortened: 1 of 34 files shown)


TahaAttari:
I believe I've addressed all the gaps previously identified, remaining gaps are:
- contained references aren't handled correctly, they should be prepended with the full canonical of the parent
- when getting the latest version of an artifact we should be using a semver sort

And I plan to address these in future PRs

TahaAttari:
superseded by https://github.com/cqframework/clinical-reasoning/pull/502

PR #505, opened 2024-08-19
Refactor how Questionnaires are generated during the $apply operation

### LINKED ISSUE #472: Abstract dependency tracing from PackageVisitor and create a DataRequirementsVisitor
$data-requirements and $package are very similar operations with different results.  We should be reusing what logic we can to make these operations available to more resource types.

### LINKED ISSUE #481: $extract returns error for decimal values
$extract-Operation error on observation based extraction for decimal types:

HAPI-0389: Failed to call access method: java.lang.Error: Not the right type for Observation.value[x]: decimal

Questionnaire and Response attached. 
[Questionnaire-NumericExtract.json](https://github.com/user-attachments/files/15872994/Questionnaire-NumericExtract.json)
[QR.json](https://github.com/user-attachments/files/15872995/QR.json)

the following item causes the issues:
```
              {
                    "linkId": "baseDataDecimalWeight",
                    "answer": [
                        {
                            "valueDecimal": 23.5
                        }
                    ]
                }
```

(tested on HAPI FHIR 7.2.0)

### LINKED ISSUE #482: $extract disregards definitions in Questionnaire when applying definition based extraction
Definition based extraction works only if the definitions are present in the QuestionnaireResponse

I can put absolute nonsense into Questionnaire.item.definition, or nothing, but the extraction results remain the same. Even if I remove QuestionnaireResponse.questionnaire or point it to an nonexisting URL, extraction will be performed without error, if the definitions are present in the QuestionnaireResponse.

If I remove all definitions from QuestionanireResponse, I get "HAPI-0389: Failed to call access method: java.lang.NullPointerException: Cannot invoke \"String.contains(java.lang.CharSequence)\" because \"definition\" is null", regardless of whether definitions are present in the referenced Questionnaire or not.

If I remove the reference to the questionnaire or reference a non-existing questionnaire, and no definitions are present in the QuestionnaireResponse, I get "HAPI-0389: Failed to call access method: java.lang.IllegalArgumentException: Unable to retrieve Questionnaire code map for Observation based extraction"

To me it looks like Observation based extraction relies on the annotations in Questionnaire but Definition based Extraction relies on the annotations in QuestionnaireResponse.

IMO $extract should always use the annotations from Questionnaire as first line and only use definitions from QuestionnaireResponse as a fallback, if no Questionnaire is referenced or the referenced Questionnaire cannot be found (AND produce a warning in the latter case).

see also: https://chat.fhir.org/#narrow/stream/179167-hapi/topic/SDC-Extraction.20capabilities.20in.20HAPI.20Starter

barhodes:
- The $apply operation will now always generate a Questionnaire from any action.input elements that reference a Profile. A populated QuestionnaireResponse will be returned with the generated Questionnaire contained within it.
- Generated Questionnaire items from StructureDefinitions that reference a CPG Feature Expression will now have an SDC Questionnaire Item Population Context extension with the expression.  The expression name will be set to the name of the StructureDefinition.
- Added support for the [SDC Questionnaire Launch Context](https://hl7.org/fhir/uv/sdc/STU3/StructureDefinition-sdc-questionnaire-launchContext.html) extension in the $populate operation.
- $populate now correctly expects the subject parameter to be a Reference rather than a String.
- $populate will use "definition-based-population" with Item Population Context when the items have definitions.
- $extract now uses the Questionnaire item for definition based extraction, the QuestionnaireResponse definition will only be used if the Questionnaire cannot be found by the server.
- $extract now handles answers of type decimal or integer with a [Questionnaire Unit](https://hl7.org/fhir/extensions/5.1.0/StructureDefinition-questionnaire-unit.html) extension correctly.
- $extract now handles multiple resources being returned by the [Item Extraction Context](https://hl7.org/fhir/uv/sdc/STU3/StructureDefinition-sdc-questionnaire-itemExtractionContext.html) expression.
- Added EvaluateProcessor for the Library/$evaluate operation.
- Fixed operation processors not using the proxy repository when endpoints are passed in.
- Added DataRequirementsVisitor and Processor for the $data-requirements operation.
- $data-requirements operation added for the following resources: Library, PlanDefinition, Questionnaire, ValueSet

Closes #472 
Closes #481 
Closes #482


cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/common/IQuestionnaireRequest.java
```diff
@@ -1,23 +1,76 @@
...
+                        .orElse(null);
+                var exists =
+                        getQuestionnaireAdapter()
+                                .getExtensionsByUrl(Constants.SDC_QUESTIONNAIRE_LAUNCH_CONTEXT)
+                                .stream()
+                                .anyMatch(lc -> lc.getExtension().stream()
+                                        .map(c -> (IBaseExtension<?, ?>) c)
+                                        .anyMatch(c -> c.getUrl().equals("name")
+                                                && resolvePathString(c.getValue(), "code")
+                                                        .equals(code)));
```
TahaAttari, line 43: possible false positive here if `code = null`?

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/visitor/DataRequirementsVisitor.java
```diff
@@ -0,0 +1,246 @@
...
+            var reference = adapter.hasVersion()
+                    ? adapter.getUrl().concat(String.format("|%s", adapter.getVersion()))
+                    : adapter.getUrl();
+            boolean addArtifact = relatedArtifacts.stream()
+                    .noneMatch(ra -> KnowledgeArtifactAdapter.getRelatedArtifactReference(ra)
+                            .equals(reference));
+            if (addArtifact) {
+                relatedArtifacts.add(KnowledgeArtifactAdapter.newRelatedArtifact(
+                        fhirVersion, "depends-on", reference, adapter.getDescriptor()));
+            }
```
TahaAttari, line 212: In the event that an unversioned `relatedArtifact` reference already exists, this would result in a duplicate `relatedArtifact` with a reference?

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/common/IQuestionnaireRequest.java
```diff
@@ -1,23 +1,76 @@
...
+                        .findFirst()
+                        .orElse(null);
+                var exists =
+                        getQuestionnaireAdapter()
+                                .getExtensionsByUrl(Constants.SDC_QUESTIONNAIRE_LAUNCH_CONTEXT)
+                                .stream()
+                                .anyMatch(lc -> lc.getExtension().stream()
+                                        .map(c -> (IBaseExtension<?, ?>) c)
+                                        .anyMatch(c -> c.getUrl().equals("name")
+                                                && resolvePathString(c.getValue(), "code")
```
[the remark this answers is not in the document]
barhodes: Indeed!  Thank's for catching that!

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/visitor/DataRequirementsVisitor.java
TahaAttari: resolving as per our discussion 👍🏾 (TLDR : not possible for there to be unversioned `relatedArtifact`s)

Capt-Mac:
decent handful of issues flagged, don't see anything majorly critical, but I'd have a look through https://sonarcloud.io/project/issues?id=cqframework_clinical-reasoning&pullRequest=505&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true before merging

Capt-Mac (approved):

barhodes:
> decent handful of issues flagged, don't see anything majorly critical, but I'd have a look through https://sonarcloud.io/project/issues?id=cqframework_clinical-reasoning&pullRequest=505&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true before merging

Fixed all of them but the too many parameters in methods.  Those will have to wait for another refactor.

PR #509, opened 2024-08-29
ValueSet expansion should not fail on an empty Grouping ValueSet

TahaAttari:
[APHL-1156] - ValueSet expansion should not fail on an empty grouper


sliver007 (approved):
Looks good

PR #510, opened 2024-09-03
Add interfaces to R4MeasureService and R4MultiMeasureService.

lukedegruchy:


Capt-Mac (commented):
fix: our other interfaces are in measure/common path, similar to MeasureProcessor interface.

Capt-Mac (commented):
question: Any reason we are doing it at R4 level only?

Capt-Mac (approved):

PR #511, opened 2024-09-05
APHL-1023-withdraw_draft_programs

ibaisi:


cqf-fhir-utility/src/main/java/org/opencds/cqf/fhir/utility/visitor/WithdrawVisitor.java
```diff
@@ -0,0 +1,66 @@
...
+        var transactionBundle = BundleHelper.newBundle(fhirVersion, null, "transaction");
+        for (var artifact : resourcesToUpdate) {
+            var entry = PackageHelper.deleteEntry(artifact);
+            BundleHelper.addEntry(transactionBundle, entry);
+        }
+
+        return repository.transaction(transactionBundle);
+    }
+
+    private List<IDomainResource> gatherComposedOfChildren(
```
sliver007, line 44: NITPICK: Calling this getComponents might be more appropriate and consistent with what we've done elsewhere. "Component" is the concept that implies a child via "composed-of" reference.

sliver007 (approved):
Looks good to me.

cqf-fhir-utility/src/main/java/org/opencds/cqf/fhir/utility/visitor/WithdrawVisitor.java
ibaisi: fixed :)

PR #526, opened 2024-09-23
measure basis compliance rule

### LINKED ISSUE #525: care-gap boolean-basis Measures only
# Description

Currently care-gaps allows for any Measure resource of scoring type 'ratio' or 'proportion' to processed by the 'care-gap' population. The problem is that the basis of each Measure also needs to be restricted in order to have the 'CareGapStatusEvaluator' algorithm appropriately process the results of a Measure Report.

## issue

When a Measure is a basis of 'boolean' it processes per subject whether or not they meet the criteria for care-gaps by looking at the improvementNotation, numerator, and denominator values. This means a subject is either "inDenominator" && "inNumerator" or not. This allows for a simple algorithm to say "open-gap" or "closed-gap".

The issue with 'resource' based Measures is that they can have criteria expressions for each population that produce multiple counts per population code for a specific subject/patient. This means that a patient could produce results that apply to all care-gap status critieria if interpreted per resource, or could gloss over the purpose of the Measure by just seeing if there is a positive value in the 'Numerator'.

Ideally a Care-gap useable Measure would be authored in a way that is Patient-care focused, where they either received recommended care, or did not.

### Example Resource Measure

% of Encounters that received a specific test.

- initial-population: [Encounter-1][Encounter-2][Encounter-3][Encounter-4]
- denominator (opportunities for test Encounters): [Encounter-1][Encounter-2][Encounter-3]
- numerator (received test Encounters): [Encounter-1]

if evaluated on a Patient level per resource.
- [Encounter 4]- resource is 'not-applicable' to Measure
- [Encounter 3] & [Encounter 2] - resource is 'open-gap'
- [Encounter 1] - resource is 'closed-gap'

Therefore each resource would receive an applicable status, on a patient-level they would have 3 different care-gap status levels for the Measure, which is difficult to interpret.

current algorithm, if evaluated on Patient level and positive count per population
- inDenominator = true
- inNumerator = true

Therefore 'closed-gap' is the outcome, but that does not communicate the actual purpose of the measure, that care was met for 33% of Encounters. If a provider is meant to action this, then this interpretation fails to meaningfully derive a care-gap to resolve.

## proposed solution

There is no one-size fits all interpretation of 'resource' based measures that allows for an algorithm to produce meaningful results in a standardized way, therefore we should not allow these types of Measures to be processed.

This request would restrict $care-gaps to only "boolean" basis Measures, to avoid a non-standardized solution. 

Any existing Measures that look to leverage care-gaps, but are currently a 'resource' based Measure, will require a different implementation that is 'boolean' based in order to work.

This still allows for future implementations to solution 'resource' based measures, but until then will throw an error if requested.

Capt-Mac:
closes #525

### DIFF
--- a/cqf-fhir-cr/src/test/java/org/opencds/cqf/fhir/cr/measure/r4/R4CareGapsTest.java
+++ b/cqf-fhir-cr/src/test/java/org/opencds/cqf/fhir/cr/measure/r4/R4CareGapsTest.java
@@ -507,4 +507,28 @@ class R4CareGapsTest {
                             "MeasureScoring type: Continuous Variable, is not an accepted Type for care-gaps service"));
         }
     }
+
+    // MinimalProportionResourceBasisSingleGroup
+    @Test
+    void MinimalProportionResourceBasisSingleGroup_Subject() {
+        try {
+            GIVEN_REPO
+                    .when()
+                    .subject("Patient/female-1988")
+                    .periodStart("2024-01-01")
+                    .periodEnd("2024-12-31")
+                    .measureIds("MinimalProportionResourceBasisSingleGroup")
+                    .statuses("closed-gap")
+                    .statuses("open-gap")
+                    .getCareGapsReport()
+                    .then()
+                    .hasBundleCount(1);
+            fail("resource based measures should fail");
+        } catch (IllegalArgumentException e) {
+            Assertions.assertTrue(
+                    e.getMessage()
+                            .contains(
+                                    "CareGaps can't process Measure: MinimalProportionResourceBasisSingleGroup, it is not Boolean basis"));
+        }
+    }
 }
--- a/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4CareGapsProcessor.java
+++ b/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4CareGapsProcessor.java
@@ -154,6 +154,15 @@ public class R4CareGapsProcessor {
         for (Measure measure : measures) {
             checkMeasureScoringType(measure);
             checkMeasureImprovementNotation(measure);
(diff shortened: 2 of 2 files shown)


cqf-fhir-cr/src/test/java/org/opencds/cqf/fhir/cr/measure/r4/R4CareGapsTest.java
```diff
@@ -507,4 +507,28 @@ void ContinuousVariable_ScoringTypeError() {
                             "MeasureScoring type: Continuous Variable, is not an accepted Type for care-gaps service"));
         }
     }
+
+    // MinimalProportionResourceBasisSingleGroup
+    @Test
+    void MinimalProportionResourceBasisSingleGroup_Subject() {
```
lukedegruchy, line 513: Question:  is the below the part that makes it a non-boolean basis?

```json
  "extension": [ {
    "url": "http://hl7.org/fhir/us/cqfmeasures/StructureDefinition/cqfm-populationBasis",
    "valueCode": "encounter"
  },
```

lukedegruchy (approved):

cqf-fhir-cr/src/test/java/org/opencds/cqf/fhir/cr/measure/r4/R4CareGapsTest.java
Capt-Mac: if "basis" is not explicitly a part of a measure profile, so the extension assists in identifying this

1. no extension => assumes "boolean" basis
2. extension with valueCode = "boolean" => "boolean" basis
3. extension with valueCode != "boolean" => "resource" basis or non-boolean

this communicates if CQL results are a collection of resources, or simply a boolean

on a Measure report, this communicates that population.count, is an accumulation of resources or accumulation of boolean=true for the subject

Issue #528, opened 2024-09-24
Ensure measurement periods respect configured timezone or UTC default


lukedegruchy:
Currently, evaluated measure periods are computing using the server's timezone.

This needs to change to take a configured timezone in EvaluationSettings.

If no timezone is set in EvaluationSettings, it must default to UTC.

All computed intervals and periods must take this timezone behaviour into account.

PR #529, opened 2024-09-24
Add new EvaluationSettings for client timezone.  Default to UTC if this is not set.

lukedegruchy:
- Add new EvaluationSettings for client timezone.  Default to UTC if this is not set.
- Ensure all Interval and Period calculation uses this timezone.
- Ensure InMemoryFhirRepository doesn't throw when calling transaction()

Closes https://github.com/cqframework/clinical-reasoning/issues/528


lukedegruchy:
Superseded by: https://github.com/cqframework/clinical-reasoning/pull/538

PR #540, opened 2024-10-02
535 measure evaluation should respect improvement notation specified on group

### LINKED ISSUE #535: Measure Evaluation should respect 'Improvement Notation' specified on Group
# Description

Measure resources have the ability to specify 'improvement notation' on the Measure level through the field Measure.improvementNotation. This influences how Measure.Group component population values are meant to be read.

if "increase" for a proportion/ratio measure:

- numerator=0
- denominator=10
- measureScore=0%

this implies "numerator" count should 'increase' to improve score. 

if "decrease" for a proportion/ratio measure:

- numerator=0
- denominator=10
- measureScore=100%

this implies "numerator" count should 'decrease' to improve score.

# issue

Measure scores can also define scoringType definitions on a Measure.group level, which means that one Measure.group could have a different scenario and scoring algorithm defined. This means that if there was a need to have "improvement notation" interpreted differently per group, like scoring, then users should have a way to communicate that.

# solution

Measure.group will accept an extension specifying improvementNotation, if it is different than what is specified on the Measure resource level. Similar to group scoring definitions, if 'improvementNotation" is not specified on the group via an extension, then it will assume the Measure defined 'improvement notation' value. If no 'improvement notation' is specified on the resource, the assumed 'improvement notation' behavior is 'increase'.

1. Check Group for 'improvement notation'
2. Check Measure for 'improvement notation'
3. 1 & 2 missing, Default to 'increase' for 'improvement notation'

Extension format

```
"extension": [ {
        "url": "http://hl7.org/fhir/us/cqfmeasures/StructureDefinition/cqfm-improvementNotation",
        "valueCodeableConcept": {
          "coding": [
            {
              "system": "http://terminology.hl7.org/CodeSystem/measure-improvement-notation",
              "code": "decrease",
              "display": "decrease"
            }
          ]
        }
      } ]
```

Capt-Mac:
closes #535


cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/dstu3/Dstu3MeasureDefBuilder.java
```diff
@@ -101,9 +101,11 @@ public MeasureDef build(Measure measure) {
                     group.getId(),
                     null, // No code on group in dstu3
                     stratifiers,
-                    populations);
+                    populations,
+                    groupMeasureScoringCode,
+                    groupIsIncreaseImprovementNotation);
             groups.add(groupDef);
-            groupMeasureScoring.put(groupDef, groupMeasureScoringCode);
+            // groupMeasureScoring.put(groupDef, groupMeasureScoringCode);
```
barhodes, line 108: Remove instead of comment?

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/common/MeasureEvaluator.java
```diff
@@ -597,7 +596,7 @@ protected void evaluateCohort(
         if (initialPopulation == null) {
             throw new NullPointerException("`" + INITIALPOPULATION.getDisplay()
                     + "` is a required Population Definition for Measure Scoring Type: "
-                    + measureDef.scoring().get(groupDef).toCode());
```
barhodes, line 600: The measureDef parameter is no longer used at all in this method.  Consider removing it or add a comment about why it's there.

barhodes (approved):

PR #543, opened 2024-10-04
542 care gaps non document mode

### LINKED ISSUE #542: care-gaps non-document mode
# Description

The issue with the "document" structure of the 'care-gaps' service is that the level of detail isn't always required for every use case of the operation. There are use cases where a user just needs to know who has a specific "gap-status" and have minimal evidence support of a 'Measure Report' resource that determined the "gap" state for the subject.

This ticket is to propose that we introduce a "non-document" mode of "care-gaps" service that allows for minimal response structure for downstream use cases of the service that don't need the large amount of detail.

The current response body of the operation into Parameters resource:

- Parameter
- - Bundle -> 1 Per-Subject 
- - - BundleEntries
- - - - Composition
- - - - DetectedIssue(s) -> 1 per Measure
- - - - MeasureReport(s) -> 1 per Measure
- - - - EvaluatedResources -> Many Resources from MeasureReport.evaluatedResource values
- - - - Section Author Resource (Organization)
- - - - Reporter Resource (Organization)

# Proposed Solution

Update the 'care-gaps' operation to have an additional optional parameter that controls whether the service returns "non-document" response structure or "document" response structure. The parameter would be "isDocumentMode", which would default to "true" to return current 'document' structure unless requested otherwise, so not to interrupt current users of operation.

I also propose that with this "non-document" mode we add additional context to the "detectedIssue" to surface metaData that would make it more useful for downstream processes to identify Measure related care-gaps. Current Detected Issue does not provide enough detail here and requires the Measure Report resource to surface.

- **(MeasureReport.period->DetectedIssue.identifiedPeriod):** this indicates that the Measure 'Detected Issue' is relevant to the 'measurement period' of the Measure Report and doesn't necessarily indicate relevance outside of that period.
- **(Measure.id ->DetectedIssue.implicated):** this exposes the resource that identified the 'DetectedIssue', exposing this will allow for any persisted 'DetectedIssue' resources to be searchable by related Measure resource. 

Contain 'Measure Report' resource in 'Detected Issue' instead of two separate resources to minimize qty of resources that would require persistence. This allows for an all in one resource to identify 'care-gap' status and evidence for why they have that status.

- **(Measure Report -> DetectedIssue.contained):** this will contain the entire Measure Report resource within the 'Detected Issue'
- **(DetectedIssue.contained.MeasureReport.id -> DetectedIssue.evidence.detail):** this would place a '#' prefixed reference to the contained 'Measure Report' resource. 

New structure would look like:

 - Parameter
- - Bundle -> 1 Per-Subject 
- - - BundleEntries
- - - - DetectedIssue(s) -> 1 per Measure

Capt-Mac:
closes #542


cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4CareGapsBundleBuilder.java
```diff
@@ -98,7 +98,8 @@ public List<Parameters.ParametersParameterComponent> makePatientBundles(
             @Nullable ZonedDateTime periodEnd,
             List<String> subjects,
             List<String> statuses,
-            List<IdType> measureIds) {
+            List<IdType> measureIds,
+            boolean isDocumentMode) {
```
lukedegruchy, line 102: Why not expose both makePatientBundle() methods as makePatientBundleNonDocumentMode() and makePatientBundleDocumentMode(), make the other methods private, and only pass the boolean internally?   This will be a clearer contract for API consumers.

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4CareGapsBundleBuilder.java
```diff
@@ -215,18 +226,19 @@ private Composition.SectionComponent getSection(
...
-                .withType(CARE_GAPS_CODES.get("http://loinc.org/96315-7"))
-                .withStatus(Composition.CompositionStatus.FINAL.toString())
-                .withTitle("Care Gap Report for " + Ids.simplePart(patient))
-                .withSubject(Ids.simple(patient))
-                .withAuthor(Ids.simple(configuredResources.get("care_gaps_composition_section_author")))
-                // .withCustodian(organization) // TODO: Optional: identifies the organization
-                // who is responsible for ongoing maintenance of and accessing to this gaps in
-                // care report. Add as a setting and optionally read if it's there.
-                .build();
+    private Composition getComposition(Patient patient, boolean isDocumentMode) {
```
lukedegruchy, line 229: Is a null composition a valid use case?  If so, please make this method as @Nullable.

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4CareGapsBundleBuilder.java
```diff
@@ -262,6 +281,14 @@ private DetectedIssue getDetectedIssue(
             groupIdExt.setValue(new StringType(measureReportGroupId));
             detectedIssue.setExtension(Collections.singletonList(groupIdExt));
         }
+        if (!isDocumentMode) {
```
lukedegruchy, line 284: Could we conceivably have a third mode in the future, in addition to non document and document modes?  If so, this may be best captured by an enum, not a boolean.

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4CareGapsProcessor.java
```diff
@@ -68,7 +70,8 @@ public Parameters getCareGapsReport(
             List<String> statuses,
             List<IdType> measureIds,
             List<String> measureIdentifiers,
-            List<CanonicalType> measureUrls) {
+            List<CanonicalType> measureUrls,
+            Boolean isDocumentMode) {
```
lukedegruchy, line 74: Under what circumstances will a caller pass null instead of true or false?

lukedegruchy (approved):

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4CareGapsBundleBuilder.java
Capt-Mac: attempted this initially, but ended up having too much duplicate code, this seemed more efficient

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4CareGapsProcessor.java
Capt-Mac: when a user is using old $care-gaps request where parameter is not specified

I have instead passed this as boolean to avoid

PR #550, opened 2024-10-08
care-gap measure parameter lift null pointer fix

Capt-Mac:


cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4CareGapsService.java
```diff
@@ -71,10 +73,24 @@ protected List<Either3<IdType, String, CanonicalType>> liftMeasureParameters(
             List<IdType> measureId, List<String> measureIdentifier, List<CanonicalType> measureUrl) {
 
         List<Either3<IdType, String, CanonicalType>> eitherList = new ArrayList<>();
-        measureId.forEach(id -> eitherList.add(Eithers.forLeft3(id)));
-        measureIdentifier.forEach(identifier -> eitherList.add(Eithers.forMiddle3(identifier)));
-        measureUrl.forEach(canonical -> eitherList.add(Eithers.forRight3(canonical)));
 
+        measureId.stream()
+                .filter(x -> x.getIdPart() != null)
+                .filter(Objects::nonNull)
```
lukedegruchy, line 78: The second filter is redundant.

Capt-Mac: hah, caught it right when i merged it

lukedegruchy (approved):

PR #555, opened 2024-10-09
Fix multiple iterations of an artifact in ReleaseVisitor

barhodes:
Added a check to prevent the same artifact from being iterated through multiple times in the ReleaseVisitor if that artifact is referenced by multiple artifacts.


sliver007 (approved):
Looks good. Thank you!

PR #556, opened 2024-10-10
Remove race condition between different evaluation periods by deleting Parameters parameter from evaluateMeasure().

lukedegruchy:


cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/utils/R4DateHelper.java
```diff
@@ -38,6 +39,8 @@ public Period buildMeasurementPeriod(Interval measurementPeriodInterval) {
 
     private DateTime convertToDateTime(ZonedDateTime zonedDateTime) {
         final OffsetDateTime offsetDateTime = zonedDateTime.toOffsetDateTime();
-        return new DateTime(offsetDateTime);
+        final DateTime convertedDateTime = new DateTime(offsetDateTime);
+        convertedDateTime.setPrecision(Precision.SECOND);
+        return convertedDateTime;
```
Capt-Mac, line 44: question: does hardcoding this value work because it always receives dateTime with seconds specified?

Capt-Mac (approved):

barhodes (approved):

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/utils/R4DateHelper.java
lukedegruchy: As discussed, we'll change the precision based on the composition of the ZonedDateTime.

PR #559, opened 2024-10-13
Refactor "Engines" class

JPercival:
* Rename overloads and reorder parameters so that they are consistent across each
* Refactor NPM logic so that it's consolidated in one place
* Refactor LibrarySourceProvider creation so that it's (mostly) in one place
* Fix duplicate FHIRLibrarySourceProvider bug


barhodes (approved):
Somewhat unrelated note on the change in R4LibraryEvaluationService.  I recently created a LibraryProcessor to handle the $package and $data-requirements operations for Library and also added an EvaluationProcessor for the $evaluate operation.  I suggest we deprecate this class in favor of the new since this one is R4 only and the new class supports any version we have support for, as well as support for the prefetch parameter.

PR #568, opened 2024-10-25
Cleanup visitor test data

Chris0296:
We have excessively large test bundles being used in the Visitor tests. They are useful for the eCR plugin to be more representative of the real VSM data with its idiosyncracies but are bloat for this project.

Trimming those bundles down to the minimum information which is sufficient to test functionality


cqf-fhir-utility/src/test/java/org/opencds/cqf/fhir/utility/visitor/dstu3/ReleaseVisitorTests.java
```diff
@@ -296,68 +296,14 @@ void visitLibraryTest() {
...
-                "http://hl7.org/fhir/us/ecr/StructureDefinition/us-ph-triggering-valueset",
-                "http://snomed.info/sct",
-                "http://hl7.org/fhir/sid/icd-10-cm",
-                "http://loinc.org",
-                "http://www.nlm.nih.gov/research/umls/rxnorm");
+                "http://ersd.aimsplatform.org/fhir/PlanDefinition/us-ecr-specification|" + existingVersion,
+                "http://ersd.aimsplatform.org/fhir/Library/rctc|" + existingVersion,
+                "http://ersd.aimsplatform.org/fhir/ValueSet/dxtc|" + existingVersion,
+                "http://cts.nlm.nih.gov/fhir/ValueSet/2.16.840.1.113762.1.4.1146.6|20210526",
+                "http://snomed.info/sct");
```
TahaAttari, line 303: since this is also a test of the dependency tracing, would it be possible to also add at least one profile dep (e.g. http://hl7.org/fhir/us/ecr/StructureDefinition/us-ph-triggering-valueset-library) and a component reference which does not have the `isOwned` extension?

PR #569, opened 2024-10-29
Fix $populate issue with missing responseItems for a null population context

barhodes:
When using the item population context extension, a responseItem should now always be returned even when there is no data to populate an answer.


cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/plandefinition/apply/ApplyProcessor.java
```diff
@@ -127,7 +127,7 @@ protected void initApply(ApplyRequest request) {
         var version = request.resolvePathString(request.getPlanDefinition(), "version");
         if (version != null) {
             var subject = request.getSubjectId().getIdPart();
-            var formatter = new SimpleDateFormat("yyyy-MM-dd-hh.mm.ssZ");
```
barhodes, line 130: Removing the timezone here to prevent issues when the server is ahead of GMT.  The timestamp is only being added to ensure uniqueness of the artifact removing it will not have adverse effects.

PR #570, opened 2024-10-29
Implement SDC 4.0 changes

### LINKED ISSUE #581: $apply does not produce output for ActivityDefinitions
I have a PlanDefinition resource that utilizes ActivityDefinition and other PlanDefinition resources in the actions. The ActivityDefinitions are targeted to create ServiceRequest resources. The PlanDefinitions seem to process ok , but the ActivityDefinitions do not appear to produce any output in the CarePlan RequestGroup.  

Here is an example of the output for an action in the RequestGroup -- 
{
     "id": "ENACT-5",
     "title": "ACT-3",
      "description": "Cognition Assessment",
      "resource": {
                        "reference": "ENACT-5-ACT-3"
       }
 }

While there is a reference to I would expect to be the ServiceRequest generated by the ActivityDefinition, there are no ServiceRequest resources generated or other resources with that id.  I also question whether references based solely on an id are valid.  I would assume that references to other contained resources in the same parent context, the CarePlan in this case, would need to follow the #{resourceType}/{resourceId} semantics, but I will add that as a different issue.

### LINKED ISSUE #582: $apply produces invalid references
In the resulting CarePlan that is produced from the $apply operation there are potentially a number of contained resources that are produced and included in the CarePlan.  These resources may also have a need to reference other contained resources in the CarePlan.  Currently, when this occurs invalid references are generated.  Below is an example of the output I received in the contained RequestGroup in the CarePlan that was generated from a PlanDefinition.  The action makes reference to another contained RequestGroup with the id EN-1-EN-1.  This should be #EN-1-EN-1.  

Per the fhir specification: _When a contained resource references its container or another contained resource within its container, it must always be a local relative reference (i.e. it must be or start with '#')_

```
"action": [
                {
                    "id": "EN-1",
                    "title": "Enrollment",
                    "description": "Initial enrollment visit for patient screening and inclusion",
                    "relatedAction": [
                        {
                            "actionId": "EN-1",
                            "relationship": "concurrent",
                            "offsetDuration": {
                                "value": 0,
                                "system": "http://unitsofmeasure.org",
                                "code": "mo"
                            }
                        }
                    ],
                    "resource": {
                        "reference": "EN-1-EN-1"
                    }
  },

```

barhodes:
Implemented SDC 4.0 changes for definition based extraction and resolved issues with Questionnaire generation and population.

Fixes #581 
Fixes #582


cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/common/ExtensionBuilders.java
```diff
@@ -108,50 +109,49 @@ public static IBaseBooleanDatatype buildBooleanType(FhirVersionEnum fhirVersion,
...
     public static <T extends IBaseExtension<?, ?>> T buildSdcLaunchContextExt(
-            FhirVersionEnum fhirVersion, String code, String resourceType) {
+            FhirVersionEnum fhirVersion, String code, List<String> resourceTypes) {
         var system = "http://hl7.org/fhir/uv/sdc/CodeSystem/launchContext";
         var display = "";
         switch (code) {
             case "patient":
                 display = "Patient";
-                resourceType = display;
+                resourceTypes = Arrays.asList(display);
```
lukedegruchy, line 118: Do these Lists need to be mutable?  If not, I suggest List.of()

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/common/ExtensionBuilders.java
```diff
@@ -108,50 +109,49 @@ public static IBaseBooleanDatatype buildBooleanType(FhirVersionEnum fhirVersion,
...
                 break;
             case "encounter":
                 display = "Encounter";
-                resourceType = display;
+                resourceTypes = Arrays.asList(display);
```
lukedegruchy, line 122: I strongly suggest you clone resourceTypes into a new variable, like resourceTypesToUse, instead of hijacking the constructor argument.

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/common/ItemValueTransformer.java
```diff
@@ -6,38 +6,50 @@
...
     private ItemValueTransformer() {}
 
-    @SuppressWarnings("rawtypes")
     public static org.hl7.fhir.dstu3.model.Type transformValueToItem(org.hl7.fhir.dstu3.model.Type value) {
-        if (value instanceof org.hl7.fhir.dstu3.model.CodeableConcept) {
-            return ((org.hl7.fhir.dstu3.model.CodeableConcept) value)
-                    .getCoding()
-                    .get(0);
+        if (value instanceof org.hl7.fhir.dstu3.model.CodeableConcept codeType) {
+            return codeType.getCoding().get(0);
```
lukedegruchy, line 11: for all calls to Collection.get(0), are we certain there's at least one element, or do we need to check first?  If there isn't, we'll get a NoSuchElementException

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/inputparameters/r4/InputParameterResolver.java
```diff
@@ -115,44 +116,36 @@ protected boolean validateContext(SDC_QUESTIONNAIRE_LAUNCH_CONTEXT_CODE code, St
...
-                throw new IllegalArgumentException(String.format("Missing content for context: %s", name));
-            }
-            var value = getValue(type, content);
-            if (!value.isEmpty()) {
-                var resource =
-                        (Resource) (value.size() == 1 ? value.get(0) : BundleHelper.newBundle(FhirVersionEnum.R4));
-                if (value.size() > 1) {
-                    value.forEach(
-                            v -> ((Bundle) resource).addEntry(new BundleEntryComponent().setResource((Resource) v)));
+            if (content != null && !content.isEmpty()) {
```
lukedegruchy, line 127: Why are we no longer throwing in the opposite case?

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/inputparameters/r4/InputParameterResolver.java
```diff
@@ -115,44 +116,36 @@ protected boolean validateContext(SDC_QUESTIONNAIRE_LAUNCH_CONTEXT_CODE code, St
...
+                var value = getValue(types, content);
+                if (!value.isEmpty()) {
+                    var cqlParameterName = name.substring(0, 1).toUpperCase().concat(name.substring(1));
```
lukedegruchy, line 130: This is weird:  are we always assuming the CQL parameter name is a single character?

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/inputparameters/r4/InputParameterResolver.java
```diff
@@ -163,15 +156,28 @@ protected List<ParametersParameterComponent> getContent(
...
                 .map(p -> {
-                    if (p.getValue() instanceof Reference) {
-                        return readRepository(
-                                fhirContext().getResourceDefinition(type).getImplementingClass(),
-                                ((Reference) p.getValue()).getReferenceElement());
+                    if (p.getValue() instanceof Reference ref) {
+                        Resource resource = null;
+                        for (var type : types) {
+                            try {
+                                resource = (Resource) readRepository(
```
lukedegruchy, line 166: Last resource wins in this case?

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/inputparameters/r5/InputParameterResolver.java
```diff
@@ -115,22 +120,18 @@ protected boolean validateContext(SDC_QUESTIONNAIRE_LAUNCH_CONTEXT_CODE code, St
...
-                    .getExtensionByUrl("type")
-                    .getValueAsPrimitive()
-                    .getValueAsString();
-            if (!validateContext(SDC_QUESTIONNAIRE_LAUNCH_CONTEXT_CODE.valueOf(name.toUpperCase()), type)) {
-                throw new IllegalArgumentException(String.format("Unsupported launch context for %s: %s", name, type));
-            }
+            var types = launchContext.getExtensionsByUrl("type").stream()
+                    .map(type -> type.getValueAsPrimitive().getValueAsString())
+                    .collect(Collectors.toList());
             var content = getContent(contexts, name);
```
lukedegruchy, line 130: validateContext no longer needed?

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/common/ItemValueTransformer.java
barhodes: We are making an assumption here, but if it's wrong it should manifest as an error and should bubble up and get caught and exposed via an OperationOutcome.

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/inputparameters/r4/InputParameterResolver.java
barhodes: According to the spec it is valid to not pass in any content.

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/questionnaire/generate/GenerateProcessor.java
```diff
@@ -65,19 +69,21 @@ public IBaseResource generate(GenerateRequest request, String id) {
...
-    @SuppressWarnings("unchecked")
-    protected <E extends ICompositeType> List<E> getElements(GenerateRequest request, IBase baseElement) {
+    protected List<IElementDefinitionAdapter> getElements(GenerateRequest request, IBase baseElement) {
+        var adapterFactory = IAdapterFactory.forFhirVersion(fhirVersion);
         return baseElement == null
                 ? null
                 : request.resolvePathList(baseElement, "element").stream()
                         .filter(e -> request.resolvePathString(e, "path").split("\\.").length > 1)
-                        .map(e -> (E) e)
+                        .map(e -> (ICompositeType) e)
```
lukedegruchy, line 85: Suggestion:   .filter(ICompositeType.class::isInstance).map(ICompositeType.class::cast)

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/questionnaire/generate/ItemGenerator.java
```diff
@@ -36,55 +41,72 @@ public class ItemGenerator {
...
         this.repository = repository;
         engine = Engines.forRepository(this.repository);
-        this.elementProcessor = elementProcessor;
         expressionProcessor = new ExpressionProcessor();
         extensionProcessor = new ExtensionProcessor();
+        elementHasCaseFeature = new ElementHasCaseFeature();
+        itemTypeIsChoice = new ItemTypeIsChoice(repository);
     }
 
     public <T extends IBaseExtension<?, ?>> Pair<IBaseBackboneElement, List<T>> generate(GenerateRequest request) {
```
lukedegruchy, line 58: Please annotate any methods that could return null with jakarta.annotation.Nullable

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/inputparameters/r4/InputParameterResolver.java
barhodes: We break from the loop after finding a resource.

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/questionnaire/generate/ItemGenerator.java
```diff
@@ -36,55 +41,72 @@ public class ItemGenerator {
...
-                            var name = request.resolvePathString(p, "name").toUpperCase();
-                            return (name.equals("PRACTITIONER"))
-                                    || request.resolvePathString(p, "use").equals("in")
-                                            && Arrays.asList(SDC_QUESTIONNAIRE_LAUNCH_CONTEXT_CODE.values()).stream()
-                                                    .map(Object::toString)
-                                                    .collect(Collectors.toList())
-                                                    .contains(name);
-                        })
+                        .filter(
+                                p -> // StringUtils.isNotBlank(request.resolvePathString(p, "name")) &&
```
lukedegruchy, line 96: Do you intend to leave both lines commented out?

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/inputparameters/r5/InputParameterResolver.java
barhodes: Yup, removed.

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/questionnaire/generate/ItemGenerator.java
```diff
@@ -106,110 +128,169 @@ protected CqfExpression getFeatureExpression(GenerateRequest request) {
...
             }
         }
+        return new ImmutablePair<>(items, valueExtensions);
+    }
+
+    protected IBaseBackboneElement processProfile(
+            GenerateRequest request, IStructureDefinitionAdapter profileAdapter, String childLinkId) {
+        // generate a new group item from the profile
+        // if the generated group item has only 1 child return it
+        throw new NotImplementedException("Nested profile generation is not currently supported.");
```
lukedegruchy, line 164: Do we expect this result to ultimately result in a 500 result code for the REST call?  If so, we should probably throw ca.uhn.fhir.rest.server.exceptions.InternalErrorException

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/library/evaluate/EvaluateRequest.java
```diff
@@ -72,6 +73,12 @@ public String getOperationName() {
         return "evaluate";
     }
 
+    @Override
+    public IBase getContext() {
+        // The is used for FHIRPath evaluation which the $evaluate operation does not support
```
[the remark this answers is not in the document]
barhodes: That context, if it has a value, ends up getting passed in as the %context parameter into any evaluation calls.  While $evaluate isn't setup do FHIRPath, throwing here would cause all evaluation to fail, which we don't want.

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/questionnaire/generate/ItemGenerator.java
barhodes: Yup, cleaned up.

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/questionnaire/generate/ItemGenerator.java
barhodes: Negative, this will result in a display item being created with the error message.

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/questionnaireresponse/extract/ProcessDefinitionItem.java
```diff
@@ -39,275 +53,579 @@ public ProcessDefinitionItem(ExpressionProcessor expressionProcessor) {
...
-        if (authorPath != null) {
-            var authorValue = request.resolvePath(request.getQuestionnaireResponse(), "author");
-            if (authorValue != null) {
-                request.getModelResolver().setValue(resource, authorPath, authorValue);
+    private ImmutablePair<String, IStructureDefinitionAdapter> getPathAdapter(
+            ExtractRequest request, Optional<IStructureDefinitionAdapter> profile, String definition) {
+        var split = definition.split("#");
+        var canonical = split[0];
+        var id = split[1];
+        // TODO: check if profile url matches canonical
```
lukedegruchy, line 247: Do you plan to resolve this TODO?

barhodes: Eventually, yes.  But it's not a priority and there are still some outstanding questions I have around that.  And we likely will have to deal with multiple definition values in the near future so this code will be revisited then.

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/questionnaireresponse/extract/ProcessDefinitionItem.java
```diff
@@ -39,275 +53,579 @@ public ProcessDefinitionItem(ExpressionProcessor expressionProcessor) {
...
+
+    protected void processRepeatingWithNested(
+            ExtractRequest request,
+            IBase parent,
+            boolean isNestedRepeating,
+            List<IBaseBackboneElement> answers,
+            String[] identifiers,
+            HashMap<String, BaseRuntimeChildDefinition> propertyDefs,
+            IStructureDefinitionAdapter profile) {
+        var parentProperty = identifiers[0];
```
lukedegruchy, line 415: What if identifiers is empty?

barhodes: Identifiers originates as a split by '.' on the path.  If we don't have any path at all we won't get this far so identifiers will always have at least 1 element.

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/questionnaireresponse/extract/ResponseBundle.java
```diff
@@ -3,68 +3,32 @@
...
-
-            var entry = new org.hl7.fhir.dstu3.model.Bundle.BundleEntryComponent();
-            entry.setResource((org.hl7.fhir.dstu3.model.Resource) resource);
-            entry.setRequest(entryRequest);
-            newBundle.addEntry(entry);
-        });
-
-        return newBundle;
-    }
+    private ResponseBundle() {}
```
lukedegruchy, line 8: No more dstu3 bundle?

cqf-fhir-cr/src/test/java/org/opencds/cqf/fhir/cr/common/ExtensionBuildersTests.java
```diff
@@ -56,8 +57,8 @@ void testBuildSdcLaunchContextExt() {
                 IllegalArgumentException.class,
                 () -> ExtensionBuilders.buildSdcLaunchContextExt(FhirVersionEnum.R4, "subject"));
 
-        var r4Ext = (org.hl7.fhir.r4.model.Extension)
-                ExtensionBuilders.buildSdcLaunchContextExt(FhirVersionEnum.R4, "user", "PractitionerRole");
+        var r4Ext = (org.hl7.fhir.r4.model.Extension) ExtensionBuilders.buildSdcLaunchContextExt(
+                FhirVersionEnum.R4, "user", Collections.singletonList("PractitionerRole"));
```
lukedegruchy, line 61: Suggest List.of() instead

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/questionnaireresponse/extract/ResponseBundle.java
barhodes: Correct.  SDC targets R4 and no one is using this with DSTU3 so no reason to keep trying to support it.

cqf-fhir-cr/src/test/java/org/opencds/cqf/fhir/cr/questionnaire/generate/ItemGeneratorTests.java
```diff
@@ -83,34 +88,38 @@ void generateItemForElementWithChildren() {
...
+    void generateItemForProfileWithExtensionSlices() {
         given().repositoryFor(fhirContextR4, "r4")
                 .when()
-                .profileId(Ids.newId(fhirContextR4, "LaunchContexts"))
+                .profileId(Ids.newId(fhirContextR4, "NzPatient"))
                 .then()
-                .hasItemCount(3)
-                .hasLaunchContextExtension(5);
+                .hasItemCount(11)
+                // .hasLaunchContextExtension(5)
```
lukedegruchy, line 122: Do you plan to leave this commented out?

barhodes: Removed

lukedegruchy (approved):

PR #572, opened 2024-10-30
Update Delete operation to include deletion of Artifact Comments

Chris0296:
Delete Operation must also delete the Basic resources which represent the comments/approvals of the specified artifact

Spotless also formatted unrelated file.


sliver007 (approved):
Looks good.

PR #584, opened 2024-11-08
get latest should be last not first element

TahaAttari:


sliver007 (approved):
Looks good. Check passed.

Issue #592, opened 2024-11-18
Measure Evaluation reportType behavior


Capt-Mac:
# Problem

When user evaluates with `$evaluate-measure` operation the expected default behavior of `reportType` parameter appears to be incorrectly setting the default value when subject parameter is not populated.

# Steps to Reproduce

execute Measure/{id}/$evaluate-measure (no subject, no reportType)

# Observed behavior

`Individual` Measure Report is created

# Expected behavior

`Summary` Measure Report is created

# acceptance criteria

per https://hl7.org/fhir/R4/measure-operation-evaluate-measure.html

populated subject parameter with no reportType should default to `subject` evalType and produce `Individual` Measure Report

non-populated subject parameter with no reportType should default to `population` evalType and produce `Summary` Measure Report

Any populated `reportType` parameter should be used instead of default behavior

See below table for accepted combinations

| subject Parameter | ReportType Parameter | EvalType Used | Measure Report Type | Note                                                 |
| ----------------- | -------------------- | ------------- | ------------------- | ---------------------------------------------------- |
| empty             | empty                | population    | summary             | default behavior, when no subject parameter provided |
| empty             | subject              | subject       | individual          |                                                      |
| empty             | population           | population    | summary             |                                                      |
| empty             | subject-list         | subject-list  | subject-list        |                                                      |
| Patient/{id}      | empty                | subject       | individual          | default behavior, when subject parameter provided    |
| Patient/{id}      | subject              | subject       | individual          |                                                      |
| Patient/{id}      | population           | population    | summary             |                                                      |
| Patient/{id}      | subject-list         | subject-list  | subject-list        |                                                      |
| Group/{id}        | empty                | subject       | individual          | default behavior, when subject parameter provided    |
| Group/{id}        | subject              | subject       | individual          |                                                      |
| Group/{id}        | population           | population    | summary             |                                                      |
| Group/{id}        | subject-list         | subject-list  | subject-list        |                                                      |
| Practitioner/{id} | empty                | subject       | individual          | default behavior, when subject parameter provided    |
| Practitioner/{id} | subject              | subject       | individual          |                                                      |
| Practitioner/{id} | population           | population    | summary             |                                                      |
| Practitioner/{id} | subject-list         | subject-list  | subject-list        |                                                      |
| Organization/{id} | empty                | subject       | individual          | default behavior, when subject parameter provided    |
| Organization/{id} | subject              | subject       | individual          |                                                      |
[...]

PR #595, opened 2024-11-22
Do not error for INDIVIDUAL reports for multiple SDEs. Centralize the logic for converting report types to eval types.

lukedegruchy:
- Do not error for INDIVIDUAL reports for multiple SDEs. 
- Centralize the logic for converting report types to eval types.
- New and updated tests

Closes https://github.com/cqframework/clinical-reasoning/issues/592


cqf-fhir-cr/src/test/java/org/opencds/cqf/fhir/cr/measure/r4/R4MeasureEvaluationTest.java
```diff
@@ -158,7 +165,8 @@ void proportionMeasureEvaluation() throws Exception {
 
         Measure measure = proportion_measure();
```
Capt-Mac, line 166: Question: do we have a test that runs this scenario:
- reportType=subject
- subject=null (or value that returns more than one patient)
and creates a MeasureReport?

lukedegruchy: New Measure test added for a Measure with multiple SDEs.

PR #597, opened 2024-11-25
CqlFhirParametersConverter EmptyList and Null support

jreyno77:
R4FhirParametersConverter does not expose the EmptyList or Null Extension conversion support.  Even so, there is a lot of crossover between it and the CqlFhirParametersConverter, and I believe it makes sense to include this functionality in the CqlFhirParametersConverter.


jreyno77:
Test Failures are unrelated to my changes

JPercival (approved):

Issue #596, opened 2024-11-25
Throw FHIR InvalidRequestException (status code 400) instead of IllegalArgumentException (500) for measure eval and care gaps


lukedegruchy:
Currently, much (or most) of CR code throws an IllegalArgumentException in the event of a request trigger by user error.

This means that upstream applications, such as jpa-server-starter will return 500 status codes instead of 400, which is not RESTful.   A 500 status code indicates a system error, which is wrong.  We want to return 400 so the user understands the failure response is due to an invalid user request.

The solution is to instead throw FHIR InvalidRequestException from measure evaluation and care gaps code, ensuring end users see 400 status codes in those situations.

PR #599, opened 2024-11-26
Ensure most cases of IllegalArgumentException for measures and caregaps are converted to FHIR InvalidRequestException

lukedegruchy:
- Convert all most cases of IllegalArgumentException for measures and caregaps are converted to FHIR InvalidRequestException where they make sense to ensure end users see 400 status codes in cases of user error
- Adjust tests

Closes https://github.com/cqframework/clinical-reasoning/issues/596


barhodes (approved):
Touching lots of code like this exposes our gaps in test coverage.  Would be great to add some tests to address some of them.

Issue #601, opened 2024-11-27
Submit Data Operation Testing Coverage


Capt-Mac:
SubmitData service was moved into Clinical Reasoning but does not have a testing class that validates that the operation is performing to specification https://www.hl7.org/fhir/R4/measure-operation-submit-data.html

The scope of this issue is to validate that submit-data implementation is matching fhir specification and has testing coverage defined to ensure that the service maintains adherence to the specification.

PR #605, opened 2024-12-02
Add tests for R4SubmitDataService and tweak docs for class itself.

lukedegruchy:
- Write new test
- Tweak javadocs

Closes:  https://github.com/cqframework/clinical-reasoning/issues/601

### DIFF
--- a/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4SubmitDataService.java
+++ b/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4SubmitDataService.java
@@ -8,6 +8,9 @@ import org.hl7.fhir.r4.model.MeasureReport;
 import org.hl7.fhir.r4.model.Resource;
 import org.opencds.cqf.fhir.api.Repository;
 
+/**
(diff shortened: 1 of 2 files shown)


cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4SubmitDataService.java
```diff
@@ -16,18 +19,22 @@ public R4SubmitDataService(Repository repository) {
         this.repository = repository;
     }
 
+    // TODO: LD: Add description for measureId parameter once it's clear what this is:
```
barhodes, line 22: It may have been intended to be the id of the Bundle?  If we don't have a need for it I would suggest removing it.

barhodes (approved):
Removing the unused argument and TODO note about it should get SonarQube passing.

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4SubmitDataService.java
lukedegruchy: Done

Issue #600, opened 2024-11-27
Measure Evaluator Testing


Capt-Mac:
# Description

MeasureEvaluator class is missing coverage for scenarios where invalid population basis are interacting.

- population to population (numerator and denominator)
- stratifier of population (population is encounter based, but stratifier is patient based)

if measure populations are defined to be of basis 'boolean' then the interaction of whether a criteria result is correctly managed requires all other populations are also 'boolean'. If the basis is resource based, like an `Encounter`, then all populations of that Measure or GroupComponent where defined need to have criteria expressions that produce a list of Encounter resources. 

This is because MeasureEvaluator Scoring algorithms manage population sets based on membership of other populations. Numerator-Exclusion membership removes matching membership from Numerator. If these had different population basis then you couldn't compare them.

see https://build.fhir.org/ig/HL7/cqf-measures/measure-conformance.html#proportion-measure-scoring for more info

# Scope

- Add logic to MeasureEvaluator to validate that criteria expressions match the population basis of the GroupDef
- Add logic to MeasureEvaluator to validate that criteria expressions for stratifiers match the population basis of the GroupDef
- Throw errors when these populations don't match
- Build out testing that validates these scenarios are covered and managed appropriately

PR #607, opened 2024-12-02
First commit:  Test group and stratifier population basis validation.

lukedegruchy:
Closes https://github.com/cqframework/clinical-reasoning/issues/600

### DIFF
--- a/cqf-fhir-cr/pom.xml
+++ b/cqf-fhir-cr/pom.xml
@@ -76,6 +76,16 @@
         <artifactId>slf4j-test</artifactId>
         <scope>test</scope>
     </dependency>
+    <!--
+    Needed to support @Language annotation for JDK 17 multi-line strings and syntax highlighting.
+    This supports SQL, JSON, etc, but unfortunately, not CQL (yet).
+    -->
+    <dependency>
+        <groupId>org.jetbrains</groupId>
+        <artifactId>annotations</artifactId>
+        <version>23.0.0</version>
+        <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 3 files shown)


cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4MultiMeasureService.java
```diff
@@ -57,9 +59,15 @@ public R4MultiMeasureService(
         subjectProvider = new R4RepositorySubjectProvider(measureEvaluationOptions.getSubjectProviderOptions());
 
         r4Processor = new R4MeasureProcessor(
-                repository, this.measureEvaluationOptions, subjectProvider, r4MeasureServiceUtils);
+                repository,
+                this.measureEvaluationOptions,
+                subjectProvider,
+                r4MeasureServiceUtils,
+                populationBasisValidator);
```
Capt-Mac, line 66: Question: do we need to pass in populationBasisValidator so soon? Wouldn't we only need to use this in MeasureEvaluator?

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4PopulationBasisValidator.java
```diff
@@ -0,0 +1,163 @@
+package org.opencds.cqf.fhir.cr.measure.r4;
+
+import ca.uhn.fhir.rest.server.exceptions.InvalidRequestException;
+import java.util.*;
+import java.util.stream.Collectors;
+import org.hl7.fhir.r4.model.*;
```
Capt-Mac, line 6: suggestion: can we do the specific import instead of *

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4MultiMeasureService.java
lukedegruchy: I pushed down instantiating of both classes as far down the constructor hierarchy as I could.

Capt-Mac (approved):

PR #621, opened 2024-12-21
 latest from Tx Server

TahaAttari:
[APHL-1326]

This ticket is for the $release operation parameter `latestFromTxServer`, it should pull the latest versions of ValueSets from the Tx server at release time if they do not have versions specified on the canonical references in their parents.

Remaining work:

- tests
- polishing
- Code Quality

[APHL-1326]: https://alphora.atlassian.net/browse/APHL-1326?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


TahaAttari:
Remaining work:

- tests
- polishing

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/visitor/ReleaseVisitor.java
```diff
@@ -177,66 +213,67 @@ private static void updateMetadata(
...
+            // ValueSets we don't own go to the Tx Server if latestFromTxServer is true
+            latest = terminologyServerClient
+                    .getResource(endpoint, preReleaseReference, this.fhirVersion())
+                    .map(r -> (IKnowledgeArtifactAdapter) createAdapterForResource(r));
+        } else {
+            // get the latest ACTIVE version only because it's NOT owned
+            latest = VisitorHelper.tryGetLatestVersionWithStatus(preReleaseReference, repository, ACTIVE);
+        }
+        return latest;
+    }
```
TahaAttari, line 271: @sliver007 Wanted to check that the logic here is correct, for components, if `latestFromTxServer` is true it should only query the Tx Server if:

1. The component is NOT owned
2. The reference does not have a version
3. The component is referencing a ValueSet

sliver007: Is there some level of 'authoritativeSource' awareness that is needed here - similar to other areas of $release?

TahaAttari: yes there is, thanks 😬

TahaAttari: we can't get the actual authoritative source though, we can only check to see if the URL on the endpoint matches the URL in the reference and then 🤞🏾

TahaAttari: As per our Slack discussion left it as is and added a comment explaining the assumption 👍🏾 (we assume the caller knows what they're doing, at worst the VSets are just not resolved)

JPercival (approved):

sliver007 (approved):
Looks good. Thank you!

PR #624, opened 2025-01-08
Add interfaces on top of R4CareGapsProcessor and R4CareGapsService

lukedegruchy:
- See https://github.com/cqframework/clinical-reasoning/pull/510 for similar work done for measure evaluation services
- Add interfaces on top of R4CareGapsProcessor and R4CareGapsService so downstream systems can use composition instead of inheritance.


lukedegruchy:
Superseded by: https://github.com/cqframework/clinical-reasoning/pull/633

PR #635, opened 2025-03-06
WIP

lukedegruchy:


barhodes (changes requested):
We should make sure each class we intend to eventually drop has a JavaDoc comment stating that.  I think it will be ok to not have any test coverage on those classes, but we should create some unit tests for the classes that will be staying.

barhodes (approved):
Test coverage of cr-hapi is low.

PR #642, opened 2025-03-13
Add Retry & Back-Off logic to VS Expansion

Chris0296:
We are seeing exceptions in VSM during Program export due to throttling from VSAC when expanding Value Sets. 

They have a hard request limit of 20/second, but it appears that throttling is also occurring due to overall volume of traffic to their server.

This retry/back-off logic is a step towards preventing VSAC throttling from impacting Program exports.

APHL-1369


barhodes (changes requested):
As discussed, create a scheduled service rather than calling thread.sleep()

Chris0296:
@barhodes 

Pushed WIP changes to use ScheduledExecutorService. This seems to be the standard approach for using it as a retry mechanism (If exception is caught in processing, schedule another execution)

The issue I am facing with this, is that any exceptions which are thrown in these subsequent scheduled executions are not propagated back to the main execution process.

Any thoughts/suggestions on this?

barhodes (approved):

PR #643, opened 2025-03-17
Fix wrong parameter type used when filtering CQL retrieves by template

barhodes:
When a CQL retrieve is filtered by a template and the engine is set to use search parameters for filtering the `_profile` searchParam was always being given a ReferenceParam.  This parameter type was changed in R5 to be a ReferenceParam but is still a UriParam in R4 and previous versions.  This fix checks the version of the FhirContext being used and sets the parameter type accordingly.

### DIFF
--- a/cqf-fhir-cql/src/main/java/org/opencds/cqf/fhir/cql/engine/retrieve/BaseRetrieveProvider.java
+++ b/cqf-fhir-cql/src/main/java/org/opencds/cqf/fhir/cql/engine/retrieve/BaseRetrieveProvider.java
@@ -46,7 +46,7 @@ import org.slf4j.LoggerFactory;
 
 public abstract class BaseRetrieveProvider implements RetrieveProvider {
     private static final Logger logger = LoggerFactory.getLogger(BaseRetrieveProvider.class);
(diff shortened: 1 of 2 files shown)


cqf-fhir-cql/src/main/java/org/opencds/cqf/fhir/cql/engine/retrieve/BaseRetrieveProvider.java
```diff
@@ -302,7 +305,10 @@ public void populateTemplateSearchParams(
         // supports the _profile
         // parameter, we should add it.
         if (this.getRetrieveSettings().getProfileMode() != PROFILE_MODE.OFF && StringUtils.isNotBlank(templateId)) {
-            searchParams.put("_profile", Collections.singletonList(new ReferenceParam(templateId)));
+            var profileParam = fhirContext.getVersion().getVersion().isOlderThan(FhirVersionEnum.R5)
```
lukedegruchy, line 308: Can you write a quick test to validate this logic?

lukedegruchy (approved):

PR #647, opened 2025-03-24
Measure population change

### LINKED ISSUE #640: Measure Report Scoring Counts
# Description

Measure Report counts for 'Proportion' scoringType measures have a complicated output with some gray area on how some population codes are displayed to users on a Measure Report. CQF-Clinical-Reasoning calculates MeasureReport counts in a way that displays calculated population counts used to calculate the measureScore/performanceRate, where the result displayed to users is how the engine code ultimately scored each population. The standard puts some expectation on users to calculate performance rate by subtracting exclusions (ex denominator-exclusion from denominator), to know the performanceRate (see below). 

The gray area on how MeasureReport Proportion measures score, where IN denominator-exception subjects will display on a MeasureReport if True. This is a problem due to how IN denominator-exception subjects are used if subject is also IN numerator. If IN numerator, the subject is not counted as a denominator-exception, and is only counted in numerator. The problem with this is, if you leave the subject InDenominator-exception and InNumerator, the Measure Report displayed values don't indicate the correct values to calculate measureScore. This is because the value showing as IN denominator-exception is showing IN, for a scenario that should be excluded from calculation.

This gray area makes it difficult to perform any aggregation on proportion type measure reports.

performanceRate = (Numerator-Numerator-Exclusion)/(Denominator - Denominator-Exclusion - Denominator-Exception)

# Proposed Solution

Update Measure Report counts to reflect standard examples that allow for above mentioned performanceRate to calculate measureScore for proportion Measures. The exception being that if subject is InNumerator & InDenominator-Exception, the Measure Report count will only show count InNumerator. 

The other scoringTypes (cohort, continuous-variable, ratio) would follow the same pattern which requires manual subtraction of any applicable exclusions to get performanceRates.

Attached is an example of the change

[measureScoring-revised.pdf](https://github.com/user-attachments/files/19215767/measureScoring-revised.pdf)

Implication on care-gaps would require an update to calculating care-gap status from MeasureReport population counts, please see the attached table for mapping of values.

[measurereport-to-care-gap.pdf](https://github.com/user-attachments/files/19217400/measurereport-to-care-gap.pdf)

Capt-Mac:
closes #640


cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/common/MeasureEvaluator.java
```diff
@@ -450,6 +450,12 @@ protected void evaluateProportion(
             // Evaluate Population Expressions
             denominator = evaluatePopulationMembership(subjectType, subjectId, denominator, evaluationResult);
             numerator = evaluatePopulationMembership(subjectType, subjectId, numerator, evaluationResult);
+            // remove denominator values not in IP
```
lukedegruchy, line 453: I'm thinking out loud here, but would it be worth wrapping an interface around this method and have 2 implementations:  the old calculation and the new calculation, in case for whatever reason we want to choose one behaviour over another in configuration?

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4MeasureReportScorer.java
```diff
@@ -15,30 +15,61 @@
...
- * according to the values populated.</p>
- *<br></br>
- * <p>The population scores within a group are each independently calculated as 'sets' and not counts.</p>
- *<br></br>
- *  <p>A person may be a member of 0, 1, or more sets.</p>
- *  <br></br>
- *  <p>The CQL returns "true" or "false" or "1" or "0" if a person is a member of a given set. It's not giving you a number to count, it's telling you whether a subject is a member of some population or not. The set math happens external to the CQL.</p>
- *<br></br>
- * <B>For example, given Patients A, B, C, D: </B>
+ * Evaluation of Measure Report Data showing raw CQL criteria results compared to resulting Measure Report.
```
lukedegruchy, line 18: Can you document the flex points in this table, namely, the old behaviour versus the new?

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4CareGapStatusEvaluator.java
```diff
@@ -23,25 +23,40 @@
...
+     * Compares MeasureReport values and calculated results for various gap statuses.
      *
-     * *`prospective-gap` status requires additional data points than just population-code values within a MeasureReport in order to determine if ‘prospective-gap’ or just ‘open-gap’.
-     * **denominator-exception: is only for ‘proportion’ scoring type.
-     * ***improvement Notation: is a Measure defined value that tells users how to view results of the Measure Report.
+     * <p>This table evaluates performance across different population criteria:
+     * Initial Population, Denominator, Denominator Exclusion, Numerator,
+     * Denominator Exception, and Numerator Exclusion.
+     *
+     * <pre>{@code
```
lukedegruchy, line 32: Can you document the flex points in this table, namely, the old behaviour versus the new, if applicable?

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4CareGapStatusEvaluator.java
```diff
@@ -80,6 +100,33 @@ private CareGapsStatusCode getGapStatus(
                     && population.getCount() == 1) {
                 inDenominator.setValue(true);
             }
+            if (population.hasCode()
```
lukedegruchy, line 103: I presume to flip this ot the old behaviour we just disable this code block?

lukedegruchy (approved):

lukedegruchy (approved):

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/common/MeasureEvaluator.java
Capt-Mac: Until there is standards language that clarifies a move towards the previous interpretation of how to display population membership on a MeasureReport, we should leave this change as is.

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4MeasureReportScorer.java
Capt-Mac: added

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4CareGapStatusEvaluator.java
Capt-Mac: yes, we would exclude consideration of exceptions and exclusions

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4MeasureReportScorer.java
```diff
@@ -118,8 +149,11 @@ protected void scoreGroup(
             case PROPORTION:
             case RATIO:
                 var score = calcProportionScore(
-                        getCountFromGroupPopulation(mrgc.getPopulation(), NUMERATOR),
-                        getCountFromGroupPopulation(mrgc.getPopulation(), DENOMINATOR));
+                        getCountFromGroupPopulation(mrgc.getPopulation(), NUMERATOR)
```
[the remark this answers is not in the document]
Capt-Mac: yes, but excluding for this mr

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4CareGapStatusEvaluator.java
Capt-Mac: documented change for release version

PR #652, opened 2025-04-08
multicontext values for cli tool

Capt-Mac:
enhancement to CLI tool to allow multiple context values to be specified for expression evaluation


cqf-fhir-cr-cli/src/main/java/org/opencds/cqf/fhir/cr/cli/command/CqlCommand.java
```diff
@@ -96,15 +96,15 @@ static class LibraryParameter {
         @Option(names = {"-e", "--expression"})
         public String[] expression;
 
-        @ArgGroup(multiplicity = "0..1", exclusive = false)
+        @ArgGroup(exclusive = false)
         public ContextParameter context;
```
brynrhodes, line 100: Isn't the multiplicity change going to affect the cardinality of the `context` field? So this should public List<ContextParameter> contexts?

Capt-Mac: It currently reuses the singular `context` for each provided `context-value`.  Wouldn't the cql file only have a single context?

brynrhodes: If that's how the command option parameters work, then okay, it just seems like if the multiplicity is on the "context" element, then that's the thing that would end up needing to be able to be list-valued? Maybe I just don't understand how the command option parameters are working.

Capt-Mac: we could add a second test where two contextNames are added which should cause a failure

JPercival: Either that or reorganize the arguments so that you specify the library/expression independently of the rest of the parameters.

PR #654, opened 2025-04-10
Use HAPI Repository interface

barhodes:


cqf-fhir-api/src/main/java/org/opencds/cqf/fhir/api/Repository.java
```diff
@@ -18,6 +17,7 @@
 import org.hl7.fhir.instance.model.api.IIdType;
 
 /**
+ * @deprecated
```
lukedegruchy, line 20: Where did this requirement come from?

cqf-fhir-benchmark/src/test/java/org/opencds/cqf/fhir/benchmark/measure/r4/Measure.java
```diff
@@ -3,13 +3,13 @@
 import static org.opencds.cqf.fhir.test.Resources.getResourcePath;
 
 import ca.uhn.fhir.context.FhirContext;
+import ca.uhn.fhir.repository.Repository;
```
lukedegruchy, line 6: What happens if we need to update this interface?  Will we now have to wait at least one more release cycle to propagate it to clinical-reasoning + cdr-cr?  Meaning, with a dependency sequence of hapi-fhir <- clinical-reasoning <- cdr-cr.

cqf-fhir-api/src/main/java/org/opencds/cqf/fhir/api/Repository.java
barhodes: We've been working towards moving this interface down into HAPI for awhile now.  That happened with the 8.0.0 release.

cqf-fhir-utility/src/main/java/org/opencds/cqf/fhir/utility/repository/InMemoryFhirRepository.java
```diff
@@ -145,30 +146,32 @@ public <B extends IBaseBundle, T extends IBaseResource> B search(
...
             for (var idQuery : idQueries) {
-                var idToken = (TokenParam) idQuery;
-                // Need to construct the equivalent "UnqualifiedVersionless" id that the map is
-                // indexed by. If an id has a version it won't match. Need apples-to-apples Ids types
-                var id = Ids.newId(context, resourceType.getSimpleName(), idToken.getValue());
-                var r = resourceIdMap.get(id);
-                if (r != null) {
-                    idResources.add(r);
+                for (var query : idQuery) {
+                    var idToken = (TokenParam) query;
```
lukedegruchy, line 154: Please use if/instanceof/cast here.

cqf-fhir-benchmark/src/test/java/org/opencds/cqf/fhir/benchmark/measure/r4/Measure.java
barhodes: Yes, and we already have that dependency for the models.  Ideally this interface will not be changing often, if at all.

lukedegruchy (approved):

PR #663, opened 2025-05-06
Upstream Measure Processor, ApplyScoringMembership configuration option

Capt-Mac:
This change allows for measure processor two pathways to generating Measure Reports:

1. Calculate Results, Process Results, Output in Measure Report (existing behavior)
2. Pass in pre-calculated Results, Process results, Output in Measure Report  (bypass result generation)

The core purpose of this change is to expose dQM enforcement points and report generation while allowing upstream result creation to come from alternative sources like the CQL CLI tool.

Another feature was added allows for introduction of a new configuration that disables/enables application of Measure Scoring by set membership between dependent populations (InNumerator=True requires InDenominator=True). The current behavior of enforcing set membership will be defaulted, but if desired, users could override this setting and have raw output of CQL criteria expressions result in Measure Report populations (InNumerator=True requires InDenominator=False).

This is to allow implementers the ability to have CQL enforce membership logic instead of relying on dQM scoring algorithms. If choosing an existing scoring algorithm, all other standard behavior is still enforced.


JPercival (approved):

JPercival (approved):

brynrhodes (approved):

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/common/MeasureProcessorUtils.java
```diff
@@ -0,0 +1,425 @@
...
+     * @param measureEvalType the type of evaluation algorithm to apply to Criteria results
+     * @param applyScoring whether Measure Evaluator will apply set membership per measure scoring algorithm
+     * @param populationBasisValidator the validator class to use for checking consistency of results
+     */
+    public void processResults(
+            Map<String, EvaluationResult> results,
+            MeasureDef measureDef,
+            @Nonnull MeasureEvalType measureEvalType,
+            boolean applyScoring,
+            PopulationBasisValidator populationBasisValidator) {
```
lukedegruchy, line 51: Question:  why not pass this down from the constructor instead?

Capt-Mac:
current code coverage gaps will be patched by addition of CQL CLI testing that will make use of MeasureProcessor 'evaluateResults' method

PR #665, opened 2025-05-13
Add direct reference codes to manifest during $release

Chris0296:
Update $release operation to gather any direct-reference codes referenced by component measures and add them to the manifest, using the directReferenceCode extension slice on the root of the release manifest.

Functionality handles STU4/STU5+, works with cqfm-directReferenceCode (STU4) and cqf-directReferenceCode (STU5) in the Measure source, output is always cqf-directReferenceCode extensions in the Manifest because cqfm-directReferenceCode isn't allowed in a CRMIManifest.

### DIFF
--- a/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/constant/MeasureConstants.java
+++ b/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/constant/MeasureConstants.java
@@ -43,4 +43,9 @@ public class MeasureConstants {
     public static final String FHIR_ALL_TYPES_SYSTEM_URL = "http://hl7.org/fhir/fhir-types";
     public static final String POPULATION_BASIS_URL =
             "http://hl7.org/fhir/us/cqfmeasures/StructureDefinition/cqfm-populationBasis";
+
+    public static final String CQFM_DIRECT_REFERENCE_EXTENSION =
+            "http://hl7.org/fhir/us/cqfmeasures/StructureDefinition/cqfm-directReferenceCode";
+    public static final String CQF_DIRECT_REFERENCE_EXTENSION =
+            "http://hl7.org/fhir/us/cqfmeasures/StructureDefinition/cqf-directReferenceCode";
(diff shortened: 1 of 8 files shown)


Chris0296:
Copying question from the related issue here:

"Is it safe to assume that STU4 will always have 'cqfm-directReferenceCode' and STU5 will always have 'cqf-directReferenceCode' or are both extension URLs valid for both versions?"

This implementation assumes the former.

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/constant/MeasureConstants.java
```diff
@@ -43,4 +43,9 @@ private MeasureConstants() {}
     public static final String FHIR_ALL_TYPES_SYSTEM_URL = "http://hl7.org/fhir/fhir-types";
     public static final String POPULATION_BASIS_URL =
             "http://hl7.org/fhir/us/cqfmeasures/StructureDefinition/cqfm-populationBasis";
+
+    public static final String CQFM_DIRECT_REFERENCE_EXTENSION =
+            "http://hl7.org/fhir/us/cqfmeasures/StructureDefinition/cqfm-directReferenceCode";
+    public static final String CQF_DIRECT_REFERENCE_EXTENSION =
+            "http://hl7.org/fhir/us/cqfmeasures/StructureDefinition/cqf-directReferenceCode";
```
brynrhodes, line 50: Spelling is incorrect here, needs to be straight to FHIR:
```suggestion
            "http://hl7.org/fhir/StructureDefinition/cqf-directReferenceCode";
```

brynrhodes (changes requested):
Test needs fixup on the cqf- extension spelling, it's an extension pack extension, so it's URL is just http://hl7.org/fhir/StructureDefinition/cqf-directReferenceCode, (i.e. remove the us/cqfmeasures part from the canonical)

barhodes (approved):

brynrhodes (approved):

PR #666, opened 2025-05-15
Add support for applicability behavior in $apply

barhodes:
Add support for `all` and `any` applicability behavior when processing actions during the PlanDefinition/$apply operation.  

`all` - meaning the current behavior of evaluating the applicability of each child action independently. 
`any` - meaning that each child action will be evaluated in order, and the first action that returns an applicability of true will be performed, and processing of the parent action will stop.

The behavior is indicated on the action by the extension "http://hl7.org/fhir/StructureDefinition/cqf-applicabilityBehavior" with a value type of `code`.
If not specified, the current default behavior of `all` is used.


brynrhodes (approved):
Looks good, and I'll note the trackers this is relevant to:

https://jira.hl7.org/browse/FHIR-50150
https://jira.hl7.org/browse/FHIR-50490

PR #683, opened 2025-06-24
Introduce clinical-reasoning changes to support HEDIS measures.  

lukedegruchy:
- Take the substance of the changes from the jp-cli-2 branch and ensure they work with the latest CR master with, among other things, the repository changes.


cqf-fhir-cr-cli/src/test/java/org/opencds/cqf/fhir/cr/cli/NdjsonBundleExtractor.java
```diff
@@ -0,0 +1,84 @@
+package org.opencds.cqf.fhir.cr.cli;
+
```
Capt-Mac, line 2: I made this utility to unbundle patient bundles into isolated patient context format for cql CLI, but have since mainly used the python scripts instead. Suggest we remove this class

cqf-fhir-cr-cli/src/test/java/org/opencds/cqf/fhir/cr/cli/CliTest.java
```diff
@@ -510,6 +501,30 @@ void qICoreEXM124Numer() {
         assertTrue(output.contains("Numerator=true"));
     }
```
Capt-Mac, line 503: fix- can we add a test that uses measure evaluation?

PR #688, opened 2025-06-27
Optimize CQL library evaluation and consequently measure evaluation

lukedegruchy:
- This leverages the changes in CQL recently merged to master.
- Shift CQL library evaluation to outside the MeasureReport building loop
- Introduce `CompositionEvaluationResultsPerMeasure`, which serves as a combined cache of CQL evaluation results and errors, designed to maintain current permissive error handling
- Shift CqlEngine initialization to these earlier points, and pass it down to the rest of measure evaluation
- Leave DSTU3 logic largely unchanged in terms of timing and efficiency, but make use of the new functionality (partially for code coverage concerns)
- Ensure R4CollectDataService works the same as before, with CQL library evaluation done per subject if the subject list is not empty
- Take the opportunity to include some non-functional refactoring and sonar fixes, including making refraining from mutating R4MeasureProcessor and R4MeasureServiceUtils fields if we encounter a proxy scenario:  Use local variables and conditional logic instead.


cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/common/MeasureProcessorUtils.java
```diff
@@ -337,51 +339,70 @@ public Object evaluateObservationCriteria(
...
-                        subjectId,
-                        libraryEngine.getEvaluationResult(
-                                id, subjectId, null, null, null, null, null, zonedMeasurementPeriod, context));
-            } catch (Exception e) {
-                // Catch Exceptions from evaluation per subject, but allow rest of subjects to be processed (if
-                // applicable)
-                var error = "Exception for subjectId: %s, Message: %s".formatted(subjectId, e.getMessage());
-                // Capture error for MeasureReportBuilder
-                measureDef.addError(error);
-                logger.error(error, e);
```
Capt-Mac, line 380: suggestion: add comment here that says why we prioritize subject context not switching between measure evaluations.

// when subject context value changes, expression caching is dropped. Ordering all subject expression evaluation together allows for maximum caching

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4MultiMeasureService.java
```diff
@@ -146,18 +171,22 @@ protected void populationMeasureReport(
...
             String reporter) {
 
-        // one aggregated MeasureReport per Measure
         var totalMeasures = measures.size();
         for (Measure measure : measures) {
             MeasureReport measureReport;
             // evaluate each measure
             measureReport = r4Processor.evaluateMeasure(
-                    measure, periodStart, periodEnd, reportType, subjects, additionalData, parameters, evalType);
+                    measure,
```
Capt-Mac, line 182: fix: Shouldn't this be a list of Measures?

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4MultiMeasureService.java
```diff
@@ -207,6 +238,7 @@ protected void subjectMeasureReport(
                 "Evaluating individual MeasureReports for {} patients, and {} measures",
                 subjects.size(),
                 measures.size());
+
         for (Measure measure : measures) {
```
Capt-Mac, line 242: fix: we should be passing in Measures and subject here, so we process the same context value for all measures before switching to a new subject

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4MultiMeasureService.java
Capt-Mac: disregard

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/common/MeasureProcessorUtils.java
lukedegruchy: Add comment

Capt-Mac (approved):

PR #694, opened 2025-07-14
Stratifier component bug

Capt-Mac:
Measure component stratifier bug fix:
- MeasureReport Stratifiers persisted component.value as Code instead of value
- MeasureReport Stratifiers persisted component.code was missing
- updated tests that check for changed component stratifier behavior


cqf-fhir-cr/src/test/java/org/opencds/cqf/fhir/cr/measure/r4/MeasureStratifierTest.java
```diff
@@ -212,13 +212,17 @@ void cohortBooleanComponentStrat() {
                 .group("group-1")
                 .stratifierById("stratifier-1")
                 .stratumCount(2)
-                .stratumByComponentText("38")
+                .stratumByComponentCodeText("Age")
```
lukedegruchy, line 215: Is this how you expect this test to fail without the new code?  "- Error creating supplemental data search parameter. This may be due to the server not supporting transactions.
ca.uhn.fhir.rest.server.exceptions.NotImplementedOperationException: HAPI-2542: transaction is not supported by this repository"

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4MeasureReportBuilder.java
```diff
@@ -400,13 +401,20 @@ protected void buildStratifier(
...
+            // StratifierComponentDef (expression name, id) -> Map<Subject String, Expression Result>
+
             // Component Stratifier
             // one or more criteria expression defined, one set of criteria results per component specified
             // results of component stratifier are an intersection of membership to both component result sets
-
-            List<Map<String, CriteriaResult>> subjectCompValues = stratifierDef.components().stream()
-                    .map(StratifierComponentDef::getResults)
-                    .toList();
+            List<Map<String, Pair<CriteriaResult, StratifierComponentDef>>> subjectCompValues =
```
lukedegruchy, line 409: I'll be honest with you:  I'm having a really hard time following this new code.  While I'm pretty sure it does what you intend to do, it's hard to read and maintain.  I strongly suggest writing one or more classes dedicated this new functionality, wrapping the myriad of nested structures with an API with a clear set of contracts that can be easily unit tested.

cqf-fhir-cr/src/test/java/org/opencds/cqf/fhir/cr/measure/r4/MeasureStratifierTest.java
Capt-Mac: The test passes. It just shares a warning that supplementalDatasearch parameter is not useable. It has nothing to do with these code changes

JPercival (approved):

PR #707, opened 2025-08-12
Release and package operation updates

sliver007:
- Added support for crmi-version-specific manifest parameters
- Update $release implementation to only add manifest parameter entries for versionless references rather than all references
- Fixed a bug in the package operation that was resulting in the input expansion parameters being updated when they shouldn't have been


cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/visitor/PackageVisitor.java
```diff
@@ -224,7 +225,24 @@ protected void handleValueSets(IBaseBundle packagedBundle, Optional<IEndpointAda
...
+                paramsCopy = ((org.hl7.fhir.r4.model.Parameters) expansionParams).copy();
+                break;
+            case R5:
+                paramsCopy = ((org.hl7.fhir.r5.model.Parameters) expansionParams).copy();
+                break;
+            default:
+                break;
+        }
+
+        var params = (IParametersAdapter) createAdapterForResource(paramsCopy);
```
barhodes, line 245:
```suggestion
        var params = (IParametersAdapter) createAdapterForResource(createAdapterForResource(expansionParams).copy());
```

cqf-fhir-utility/src/main/java/org/opencds/cqf/fhir/utility/adapter/ILibraryAdapter.java
```diff
@@ -33,5 +35,34 @@ public interface ILibraryAdapter extends IKnowledgeArtifactAdapter {
 
     void setExpansionParameters(IBaseParameters expansionParameters);
 
-    void ensureExpansionParametersEntry(String resourceType, String canonical);
+    void ensureExpansionParametersEntry(IKnowledgeArtifactAdapter adapter, String crmiVersion);
```
barhodes, line 38: It seems like much of the implementation of this method is not version specific and logic is being duplicated across the three versions when we don't need to.  Consider making this a default and pulling the duplicate private methods down.

cqf-fhir-utility/src/main/java/org/opencds/cqf/fhir/utility/adapter/dstu3/LibraryAdapter.java
```diff
@@ -236,38 +237,34 @@ public void setExpansionParameters(IBaseParameters expansionParameters) {
...
-        // extract expansion parameters
-        var expansionParameters = getExpansionParameters();
+    public void ensureExpansionParametersEntry(IKnowledgeArtifactAdapter artifactAdapter, String crmiVersion) {
+        var expansionParametersOpt = getExpansionParameters();
+        if (expansionParametersOpt.isEmpty()) {
+            return;
+        }
 
-        String parameterName;
+        Parameters expansionParameters = (Parameters) expansionParametersOpt.get();
```
barhodes, line 246: nitpick: I much prefer to see `var` used for variable declaration when possible.

cqf-fhir-utility/src/main/java/org/opencds/cqf/fhir/utility/adapter/dstu3/LibraryAdapter.java
```diff
@@ -236,38 +237,34 @@ public void setExpansionParameters(IBaseParameters expansionParameters) {
         }
     }
 
-    @SuppressWarnings("unchecked")
-    public void ensureExpansionParametersEntry(String resourceType, String canonical) {
-        // extract expansion parameters
-        var expansionParameters = getExpansionParameters();
+    public void ensureExpansionParametersEntry(IKnowledgeArtifactAdapter artifactAdapter, String crmiVersion) {
```
barhodes, line 240: Need an @Override tag here.

barhodes (approved):

PR #724, opened 2025-08-29
Add all utilities classes as well as tests to support NPM.  Test tgzs…

lukedegruchy:
… and scenarios still need to be tweaked.


cqf-fhir-utility/src/main/java/org/opencds/cqf/fhir/utility/adapter/IAdapter.java
brynrhodes: The changes in this file seem like formatting changes? Do we really need to make these changes as part of this PR?

PR #726, opened 2025-09-02
CRMI Package Operation (R4)

c-schuler:
- Added $package operation provider and service
- Added the operation to the R4 CR config
- Added documentation
- TODO - validate params and naive expansion behavior (noted in TODOs in codebase)


c-schuler:
Created naive expansion issue: #730

brynrhodes (approved):

Issue #682, opened 2025-06-23
Align Measure/$evaluate operation with DEQM STU5 publication


brynrhodes:
The current implementations of $evaluate-measure and $evaluate-measures are almost, but not quite aligned with the reconciled and now published definition of $evaluate in DEQM STU 5.0.0:

https://hl7.org/fhir/us/davinci-deqm/STU5/OperationDefinition-evaluate.html

In particular, the new specification supports:

measureId 0..* id
measureIdentifier 0..* token
measureUrl 0..* canonical
measure 0..* string (reference)
measureResource 0..* Measure

Add support for the Measure/$evaluate operation and allow it to specify multiple measures in any of the above ways.

NOTE: Supporting the measureResource parameter would be fine to consider as a separate request

brynrhodes:
@JPercival , @barhodes , agree with the concern, but it looks like we implemented 4 out of the 5? If we want to narrow it down, I agree, and propose we would support only the measureUrl (0..* canonical) parameter?

PR #735, opened 2025-09-05
Change return type of $evaluate operation to a Parameters resource

barhodes:
Change the return type of the Measure/$evaluate operation to match the DEQM STU5 publication and always return a Parameters resource with a Bundle for each Measure/Subject being evaluated.

#682


cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4MultiMeasureService.java
```diff
@@ -107,10 +107,12 @@ public Bundle evaluate(
...
         var subjects = getSubjects(subjectProvider, subject);
 
-        // create bundle
-        Bundle bundle = new BundleBuilder<>(Bundle.class)
-                .withType(BundleType.SEARCHSET.toString())
-                .build();
+        // create parameters
+        var result = new Parameters();
+        //        // create bundle
+        //        Bundle bundle = new BundleBuilder<>(Bundle.class)
```
lukedegruchy: Did you mean to leave this commented out?

PR #745, opened 2025-09-16
Supporting intersection of records that matches each of the search criteria

epeartree:
**The issue:**
Previously, utility `Searches.SearchBuilder` did not support building search arguments that when submitted to a` IRespository.search` method would returned the intersection of resources matching each of the search criteria.  An example of intersection is:

`Patient?death-date=ge2022-01-01&death-date=lt2023-01-01`

The above searches for patients deseased between 2022-01-01 and 2023-01-01.

**What was done:**
`SearchBuilder` utility was modified to return a `Multimap<String, List<IQueryParameterType>>` instead of a `Map<String, List<IQueryParameterType>>`.  The difference is the mapping where with the `Multimap`, multiple values can be associated with a key as opposed to a `Map` where only one value per key is accepted.

**How is this modification solving the issue:**
To understand the fix, we need to understand that repositories differentiate whether to create an intersection or a union search by inspect on the format of provided key/value pairs provided as input. More specifically:

| key      | Value   | Description |
|:----------|:---------:|:---------:|
| birthdate  | List(ge2023-01-01, lt2025-01-01) | is the format of an union on SP `birthdate`.  Notice the format where each elements within a list represent an OR operation|
| birthdate  | List(List(ge2023-01-01),List(lt2025-01-01)) | is 2 lists wrapped into a list.  It is the wrapping that represent the intersection|

Therefor, by modifying the return value of method SearchBuilder#build to return `Multimap<String, List<>>`, we can provide repositories with an argument (`Map<String, List<List<>>>`) suitable to intersection searches.

### DIFF
--- a/cqf-fhir-utility/src/test/java/org/opencds/cqf/fhir/utility/BundleHelperTests.java
+++ b/cqf-fhir-utility/src/test/java/org/opencds/cqf/fhir/utility/BundleHelperTests.java
@@ -274,10 +274,9 @@ class BundleHelperTests {
     }
 
     @Test
-    void testgetBundleEntryResourceIds_whenBundleHasNoEntries_shouldReturnEmptyList(){
+    void testgetBundleEntryResourceIds_whenBundleHasNoEntries_shouldReturnEmptyList() {
         Bundle bundle = new Bundle();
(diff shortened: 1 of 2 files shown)


JPercival:
couple of small sonar warnings, plus formatting needs to be fixed.

PR #762, opened 2025-10-01
Add support for paging in valueset expansions

Chris0296:
The existing expansion logic in the TerminologyClient and ExpandRunner does not do paging to account for value set expansions with more than 1000 codes. 

This change ensures that where there are more than 1000 codes, additional pages are expanded to retrieve all codes. A potential issue here is that the VSAC in particular will just keep responding, so if you actually have an infinite valueset then VSAC will just keep going - therefore I have made the max number of pages to expand a configurable value within the TerminologyServerClientSettings.


Chris0296:
@barhodes Took a different approach than discussed, I didn't change the mutability of the lists returned from the ValueSetAdapters - instead I moved the logic for appending the expansion contains elements into the adapters.

brynrhodes (approved):

barhodes (approved):

PR #763, opened 2025-10-02
Ersd import operation migration

sliver007:
Created the following for the eRSD V2 Import operation:
* IERSDV2ImportServiceFactory
* ERSDTransformProvider
* R4ERSDTransformService
* R4ImportBundleProducer
* R4ImportBundleProducerTest with supporting input resources/files


c-schuler (approved):
Looks good. The only nit to pick is test coverage. No tests for the providers and would like to see a test for the new exception as well. Happy to merge regardless

Issue #766, opened 2025-10-07
CQL Naming enforcement with Library


Capt-Mac:
# Description

When a Library resource embeds CQL content as an artifact of 'CQL/text' it is requiring alignment of Library.id, library.name, library.url and CQL identifier (name in cql file) to match according to CQL naming conventions. The standard for cql naming says only FHIR Library resource.name, resource.url and resource.version should match what is in the cql file.

When an error does occur because CQL or Library are not named correctly, it is difficult to understand what needs to be fixed. Suggest we provide better error handling on how to fix specific scenarios

* library.url & name have characters that are not CQL compliant ('.', '-')
* cql identifier contains illegal character 
* library.url and library.name does not match embedded cql identifier

# Observed Behavior

CQL naming conventions have restrictive rules that only allow '_', letters, and digits to be present
This is the recommended regex to validate: ^[A-Z][A-Za-z0-9_]{0,254}$

FHIR resource 'id' conventions don't allow use of '_', but do allow for use of '.' & '-' as well as digits and numbers:
This is the regex check for FHIR.id: ^[A-Za-z0-9\-.]{1,64}$

Current implementation appears to force the overlap of those two rules so FHIR.id, CQL identifier, FHIR.name, and FHIR.url all match. This implies no use of '_', no use of '.', no use of '-' for any of those fields.

# Expected Behavior

Library.id  should be allowed to be different than CQL identifier specified in the CQL embedded artifact.

Library.name, library.url, library.version should match CQL identifier and version

# Example 1

Fixed implementation should allow this scenario to function correctly. 

* Allow '_' for Cql identifier
* allow fhir.id, fhir.url and fhir.name to not match cql identifier
* Enforce Cql identifier matches library.name & library.url

### CQL file

```
library Sample_Reporting version '2024.2.0'

using FHIR version '4.0.1'

```

### Library Resource

```
{
  "resourceType": "Library",
  "id": "SampleReporting-2024.2.0",
  "url": "http://sample.org/fhir/Library/Sample_Reporting",
  "version": "2024.2.0",
  "name": "Sample_Reporting"
}
```

# Example 2

Fixed implementation should block this scenario

* block use of '-' & '.' with cql identifier, library.url, and library.name
* error when cql identifier does not match library.name or library.url
* error when versions of cql and library do not match

### CQL file

```
library Sample-Reporting version '2024.2.1'

using FHIR version '4.0.1'

```

### Library Resource

```
{
  "resourceType": "Library",
[...]

PR #768, opened 2025-10-07
Validate measure library URL before CQL evaluation

lukedegruchy:
- Validate that the measure library canonical URL is coherent before constructing a VersionedIdentifier from it.
- Make use of Canonicals to do this
- Enhance Canonicals with hashCode(), equals(), and toString()
- Enhance Canonicals with system
- Refactor VersionedIdentifiers to make use of Canonicals and enhance by adding the URL's system

Closes https://github.com/cqframework/clinical-reasoning/issues/766


brynrhodes (commented):
This is quite a bit of logic to handle non-compliant libraries. Why not just require the libraries to be compliant?

PR #780, opened 2025-10-15
Package Include Patch

c-schuler:
The $package operation was excluding the outcome manifest library


c-schuler:
I did not notice the coverage went down... I will add some tests.

PR #784, opened 2025-10-16
Implement CRMI Withdraw operation (R4)

Chris0296:
Added $withdraw operation processor
Added documentation
Added Integration Test

This migration process is tracked here: https://github.com/DBCG/aphl-vsm/issues/558


cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/visitor/WithdrawVisitor.java
```diff
@@ -27,12 +32,16 @@ public IBase visit(IKnowledgeArtifactAdapter rootAdapter, IBaseParameters operat
         var resToUpdate = new ArrayList<IDomainResource>();
         resToUpdate.add(rootAdapter.get());
 
-        findArtifactCommentsToUpdate(rootAdapter.get(), fhirVersion.getFhirVersionString(), repository)
+        try {
```
Chris0296: Wrapped this in try/catch as the existing logic is throwing exceptions

PR #787, opened 2025-10-17
Implement CRMI Delete operation (R4)

Chris0296:
- Added $delete operation provider and service
- Added the operation to the R4 CR config
- Added documentation
- Added Integration Test

This migration process is tracked here: https://github.com/DBCG/aphl-vsm/issues/558


barhodes (changes requested):
Can we discuss how these operation providers are being added?  It is contrary to the established pattern we have and it also is wiring up these operations on every single resource even when we do not have any support for that resource.

barhodes (commented):
Awesome!  Looks great!  Just need to allow passing a DeleteProcessor into the LibraryProcessor.

PR #801, opened 2025-10-29
Refactor resource processor constructors

barhodes:
Change how settings and operation processors are passed into the resource processors to better facilitate future growth.


cqf-fhir-cr-hapi/src/main/java/org/opencds/cqf/fhir/cr/hapi/config/r4/DeleteOperationConfig.java
```diff
@@ -6,14 +6,17 @@
...
 import org.opencds.cqf.fhir.cr.hapi.config.ProviderLoader;
 import org.opencds.cqf.fhir.cr.hapi.config.ProviderSelector;
 import org.opencds.cqf.fhir.cr.hapi.r4.library.LibraryDeleteProvider;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
 
 @Configuration
+@Import(CrProcessorConfig.class)
```
epeartree: we're importing CrProcessorConfig in every config in this package.  i would think that CrR4Config would be a better place to put it.  U most likely have good reasons to imported in every config, reasons that i may not see but please have a look.

Issue #806, opened 2025-11-03
Paging behavior is incorrect


brynrhodes:
Terminology service implementation currently assumes paging behavior of VSAC (1000 codes)

The paging behavior should be:

if expansion.total > count(expansion.contains) then page

c-schuler:
Addressed in #811

PR #811, opened 2025-11-04
Addressing Expansion Paging Issue

c-schuler:
Previously, when the terminology server returned the entire expansion and the total exceeded the getExpansionsPerPage setting, the expansion was incorrectly duplicated total / getExpansionsPerPage times. For example, a 5,000-code expansion with a per-page setting of 1,000 would be copied five times, resulting in 25,000 entries.

Addresses Issue: #806 
Part of https://github.com/DBCG/aphl-vsm/issues/573

### DIFF
--- a/cqf-fhir-utility/src/main/java/org/opencds/cqf/fhir/utility/client/ExpandRunner.java
+++ b/cqf-fhir-utility/src/main/java/org/opencds/cqf/fhir/utility/client/ExpandRunner.java
@@ -97,7 +97,7 @@ public class ExpandRunner implements Runnable {
                 var expandedValueSetAdapter = (IValueSetAdapter) createAdapterForResource(expandedValueSet);
 
                 if (expandedValueSetAdapter.getExpansionTotal()
-                                > expandedValueSetAdapter.getExpansionContains().size()) {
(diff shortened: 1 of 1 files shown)


cqf-fhir-utility/src/main/java/org/opencds/cqf/fhir/utility/client/ExpandRunner.java
```diff
@@ -96,8 +96,10 @@ public void run() {
 
                 var expandedValueSetAdapter = (IValueSetAdapter) createAdapterForResource(expandedValueSet);
 
-                if (expandedValueSetAdapter.getExpansionTotal()
-                        > terminologyServerClientSettings.getExpansionsPerPage()) {
+                if ((expandedValueSetAdapter.getExpansionTotal()
+                                > terminologyServerClientSettings.getExpansionsPerPage())
```
brynrhodes: Why would we need to check the setting here? Either the expansion we're looking at is paged and we need to request the next page, or it's not, the setting is irrelevant, right?

brynrhodes (changes requested):

PR #813, opened 2025-11-05
Ensure non-distinct subject resources are returned from populationDefs and remove/retain logic considers subjectId.

lukedegruchy:
- Ensure PopulationDef exposes retain and remove methods that consider the subjectId being acted upon.
- Ensure that both ends of remove or retain filter by subjectId.
- Ensure that all resources returned from subjectDef include duplicates
- Ensure that population counts consider duplicates across subjects
- Refactoring and cleaning up unnecessary methods, like retainOverlaps()
- Adjust test assertions

### DIFF
--- a/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/common/MeasureEvaluator.java
+++ b/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/common/MeasureEvaluator.java
@@ -15,7 +15,6 @@ import com.google.common.collect.HashBasedTable;
 import com.google.common.collect.Sets;
 import com.google.common.collect.Table;
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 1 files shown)


cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/common/PopulationDef.java
```diff
@@ -99,50 +103,14 @@ public Map<String, Set<Object>> getSubjectResources() {
         return subjectResources;
     }
 
+    public Set<Object> getResourcesForSubject(String subjectId) {
+        return subjectResources.getOrDefault(subjectId, new HashSetForFhirResourcesAndCqlTypes<>());
```
Capt-Mac: Question: does this imply that this will remove values compared to method 'getResourcesDuplicatesAcrossSubjects'? when would we want to use this vs the other?

Capt-Mac (approved):

PR #823, opened 2025-11-13
Fix issue with CDS Hooks Discovery Service ignoring contained Libraries

barhodes:
When creating a CDS Hooks Service from a PlanDefinition, if the Effective Data Requirements extension, or the library field, references a contained Library it will now correctly fetch the library from the contained resources rather than attempting to search the repository.


cqf-fhir-cr-hapi/src/main/java/org/opencds/cqf/fhir/cr/hapi/cdshooks/discovery/CrDiscoveryService.java
```diff
@@ -81,7 +82,18 @@ public ILibraryAdapter resolvePrimaryLibrary(IPlanDefinitionAdapter planDefiniti
                     fhirVersion(), planDefinition.getLibrary().get(0));
         }
         if (canonical != null) {
-            return adapterFactory.createLibrary(SearchHelper.searchRepositoryByCanonical(repository, canonical));
+            IBaseResource library;
```
lukedegruchy: Nitpick:  declare this as final to indicate you expect the if/else block to exhaust all possible initialization code paths.

Issue #807, opened 2025-11-03
Validate expansion parameters


brynrhodes:
When we request an $expand, and we receive an expansion, we should validate that the parameters in the expansion match what we expect, specifically:

We should validate for every expansion parameter we passed, that it is present in the expansion parameters for the following expansion parameters:

* system-version
* valueset-version

It may be that the server returns "used-system-version" and "used-valueset-version", we should account for that possibility, falling back to the name of the parameter we passed.

If the expansion we receive does not meet these requirements, include a note about that in the result:

Expansion for ValueSet X did not use expected expansion parameters.

PR #845, opened 2025-12-01
Expansion parameter validation and adapter support across DSTU3/R4/R5

c-schuler:
Enhancement for issue #807 

This change implements full validation of expansion parameters returned by terminology servers during $expand, adds standardized adapter support for expansion parameter manipulation, and surfaces warnings through both the expanded ValueSet and the $package operation. Comprehensive test coverage was added for all FHIR versions (DSTU3, R4, R5).

ExpandHelper
- Added validateExpansionParameters to verify that the ValueSet.expansion.parameter entries:
  - match requested system-version and valueset-version parameters,
  - accept corresponding used-system-version and used-valueset-version,
  - detect missing or mismatched values.
- Added addExpansionWarningParameter to annotate expansions with a warning parameter whenever discrepancies are detected.
- Reworked parameter handling to use ParametersUtil.getNamedParameterValueAsString instead of FhirTerser.
- Eliminated reflection usage and simplified value extraction.
- Ensures warnings are only generated when appropriate.

PackageVisitor
- Added addExpansionWarningsToOperationOutcome to surface warning parameters as OperationOutcome.issue entries during $package.
- Ensures both cached and freshly expanded ValueSets contribute warnings to the result bundle.
- Fully version-agnostic behavior using adapter methods.

ValueSet Adapters (DSTU3, R4, R5)
- Added new helper API methods:
  - addExpansionStringParameter(String name, String value)
  - hasExpansionStringParameter(String name, String value)
- R4 and DSTU3 implementations now correctly:
  - iterate over all expansion parameters,
  - support all primitive datatypes via IPrimitiveType#getValueAsString,
  - correctly detect matching and mismatched parameter values.
  - R5 implementation delegates to built-in hasParameterValue.
  - Unifies expansion parameter handling across all versions.

Unit Test Additions

ExpandHelperTest
- Added tests for:
  - correct handling when expansion parameters match,
  - missing parameters,
  - mismatched valueset-version,
  - correct addition or absence of warning parameters.

ValueSetAdapterTest (DSTU3, R4, R5)
- Added tests for:
  - negative case (hasExpansionStringParameter returns false when missing),
  - positive case after calling addExpansionStringParameter,
  - correct underlying FHIR model mutation,
  - correct primitive value matching behavior.

General Improvements
- Updated imports (assertFalse, primitive type imports).
- Fixed failing assertions related to R5's IPrimitiveType behavior.
- Ensured version-specific model handling is thoroughly validated.

These changes ensure $expand behavior is validated consistently, warnings are visible to end users, and expansion parameter semantics behave uniformly across FHIR versions. Test coverage has been significantly expanded to ensure long-term stability.


barhodes (approved):

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/visitor/ExpandHelper.java
```diff
@@ -302,4 +316,112 @@ private void expandIncluded(
...
+
+            // Neither name has the expected value -> treat as missing/mismatched
+            return !(hasRequested || hasUsed);
+        });
+
+        // 4. If *any* requested parameter is missing or mismatched, add a single warning parameter.
+        if (mismatchOrMissing) {
+            addExpansionWarningParameter(
+                    expandedValueSet,
+                    "Expansion for ValueSet %s did not use expected expansion parameters."
```
brynrhodes: I think it would be a much more useful message if it said what expected parameters weren't used.

PR #865, opened 2025-12-15
Update to cql 4.2.0

JPercival:


cqf-fhir-cql/src/main/java/org/opencds/cqf/fhir/cql/LibraryEngine.java
```diff
@@ -59,13 +60,13 @@ public EvaluationSettings getSettings() {
         return settings;
     }
 
-    private Pair<String, Object> buildContextParameter(String patientId) {
-        Pair<String, Object> contextParameter = null;
+    private kotlin.Pair<String, Object> buildContextParameter(String patientId) {
```
lukedegruchy: Just curious:  What does kotlin.Pair offer?

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/common/ContinuousVariableObservationHandler.java
```diff
@@ -131,15 +132,15 @@ private static EvaluationResult processMeasureObservation(
             final ExpressionResult observationResult =
                     evaluateObservationCriteria(result, observationExpression, groupDef.isBooleanBasis(), context);
 
-            var quantity = convertCqlResultToQuantityDef(observationResult.value());
+            var quantity = convertCqlResultToQuantityDef(observationResult.getValue());
```
lukedegruchy: Just noting here that ContinuousVariableResourceMeasureObservationTest is failing on this branch, just not sure which changes are responsible yet.

cqf-fhir-cql/src/main/java/org/opencds/cqf/fhir/cql/LibraryEngine.java
JPercival: Nothing, it's just what the engine API expects. It doesn't use the Apache Pair type any more.

cqf-fhir-cql/src/main/java/org/opencds/cqf/fhir/cql/LibraryEngine.java
```diff
@@ -169,8 +170,16 @@ public IBaseParameters evaluateExpression(
         var engine = Engines.forRepository(repository, requestSettings, bundle);
 
         var id = new VersionedIdentifier().withId(libraryName).withVersion(libraryVersion);
-        var result =
-                engine.evaluate(id.getId(), Set.of("return"), buildContextParameter(patientId), evaluationParameters);
+
+        var paramsBuilder = new EvaluationParams.Builder();
+        paramsBuilder.setParameters(evaluationParameters);
+        paramsBuilder.setContextParameter(buildContextParameter(patientId));
+        paramsBuilder.library(id, builder -> {
```
antvaset: note that previously the library was only only referenced by name without version

PR #876, opened 2026-01-07
Strict Canonical Version Matching for $package

c-schuler:
Resolution for https://github.com/DBCG/aphl-vsm/issues/593

Changes Made

  1. BaseKnowledgeArtifactVisitor.java 
  - Added findResourceMatchingVersion() method that validates version matches for canonical references
  - When a specific version is requested but not found:
    - For $package: creates an OperationOutcome error and returns null (resource excluded from package)
    - For other operations: logs a warning and uses backward-compatible behavior
  - Updated recursiveGather() to accept an array wrapper for messages to enable on-demand creation

  2. PackageVisitor.java
  - Uses array wrapper pattern to allow messages to be created during dependency gathering
  - Only adds messages to bundle if they contain actual issues (prevents empty OperationOutcomes)

  3. PackageVisitorTests.java 
  - Added packageOperation_should_report_version_mismatch_and_exclude_resource() test
  - Updated existing packageOperation_should_fail() test to handle multiple issues in OperationOutcome

  Key Behaviors

  1. Strict version checking for $package: When a manifest specifies a versioned canonical (e.g., ValueSet|2.0.0) but only a different version exists in the repository, the resource is excluded and an error is added to the OperationOutcome
  2. Error messages: Clear, actionable messages like: "Requested version '2.0.0' for resource 'http://example.org/fhir/ValueSet/test-valueset|2.0.0' not found in repository. Resource will not be included in package."
  3. Backward compatibility: Other operations ($data-requirements, etc.) maintain existing behavior with warnings only
  4. On-demand message creation: Messages OperationOutcome is created only when needed, avoiding empty OperationOutcomes in successful packages

### DIFF
--- a/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/visitor/BaseKnowledgeArtifactVisitor.java
+++ b/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/visitor/BaseKnowledgeArtifactVisitor.java
@@ -246,7 +246,8 @@ public abstract class BaseKnowledgeArtifactVisitor implements IKnowledgeArtifact
 
         // If no version was requested, attempt to return the "latest"
         if (requestedVersion == null) {
-            return IKnowledgeArtifactAdapter.findLatestVersion(bundle).orElse(null);
+            return IKnowledgeArtifactAdapter.findLatestVersion(bundle)
+                    .orElseGet(() -> (IDomainResource) BundleHelper.getEntryResourceFirstRep(bundle));
         }
 
         // Search through all entries to find one matching the requested version


cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/visitor/BaseKnowledgeArtifactVisitor.java
```diff
@@ -218,4 +225,71 @@ private IDomainResource tryGetValueSetsFromTxServer(
...
+     *         or the first resource if no version match is found and messagesWrapper is null
+     */
+    private IDomainResource findResourceMatchingVersion(
+            IBaseBundle bundle, String canonical, IBaseOperationOutcome[] messagesWrapper) {
+        var requestedVersion = Canonicals.getVersion(canonical);
+        var requestedUrl = Canonicals.getUrl(canonical);
+
+        // If no version was requested, just return the first entry
+        if (requestedVersion == null) {
+            return (IDomainResource) BundleHelper.getEntryResourceFirstRep(bundle);
```
brynrhodes: If no version is specified, shouldn't this return the "latest"?

c-schuler: Agreed that is better. I was just maintaining the previous logic from line 164 there:

`.map(bundle -> (IDomainResource) BundleHelper.getEntryResourceFirstRep(bundle))`

brynrhodes (approved):

PR #886, opened 2026-01-11
fix: Add typeName to @OperationParam for String parameters

tigloo:
I encountered a `Failure during REST processing: ca.uhn.fhir.context.ConfigurationException: HAPI-1716: Resource class[java.lang.String] does not contain any valid HAPI-FHIR annotations` error when calling `PlanDefinition/$apply` in the HAPI FHIR JPA server starter project (using HAPI FHIR 8.6.0).

It seems that HAPI relies on either a `typeName` annotation or requires the use of typed classes such as `StringType`.

The fix adds typeName = "string" to all @OperationParam annotations that use String parameters.

Affected providers:
- r4/plandefinition/PlanDefinitionApplyProvider
- r4/activitydefinition/ActivityDefinitionApplyProvider
- r4/graphdefinition/GraphDefinitionApplyProvider
- dstu3/plandefinition/PlanDefinitionApplyProvider
- dstu3/activitydefinition/ActivityDefinitionApplyProvider

Note that I tested this only with the R4 `PlanDefinitionApplyProvider`.


barhodes:
This issue has been addressed in #914

Issue #835, opened 2025-11-19
Value Based Stratifiers that Return Multiple Values


Capt-Mac:
# Issue Desc

https://jira.hl7.org/browse/FHIR-53433

When value based stratifiers are evaluated and return multiple values, that combination of multiple values is currently seen as the "value" to stratify against, instead of each individual value of the multiple being returned being treated as a stratum to capture.

# Example Issue

Stratifier results in : {value-1}, {value-2}

Measure sees this as one value to stratify, not separate ones: 

Stratum value: {value-1}, {value-2}

# Expected behavior

Stratifier results in : {value-1}, {value-2}

Measure sees this each resulting value as stratum: 

Stratum value1: {value-1}

Stratum value2: {value-2}

# Expected Component Stratifier Behavior

Component 1 Results: {value-1}, {value-2}
Component 2 Results: {value-a}

Stratum 1: {value-1}, {value-a}

Stratum 2: {value-2}, {value-a}

PR #900, opened 2026-01-27
Implement multi-value stratifiers

lukedegruchy:
closes https://github.com/cqframework/clinical-reasoning/issues/835
- Add Claude-generated logic to support multi-value stratifiers
- Add Claude-generated test for boolean basis and Encounter basis multi-value stratifiers
- Refactoring of stratifier-centric code to check for criteria stratifier instead of component stratifier
- Eliminate all references to the component stratifier as well as the method in StratifierDef
- More refactoring of stratifier code to get rid of dead code
- Resolve a number of sonar issues
- More testing enhancements

### DIFF
--- a/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4StratifierBuilder.java
+++ b/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4StratifierBuilder.java
@@ -70,11 +70,11 @@ class R4StratifierBuilder {
         if (stratifierDef.isCriteriaStratifier()) {
             criteriaStratifier(bc, stratifierDef, reportStratifier, populations, groupDef);
         } else {
-            nonCriteriaStratifier(bc, stratifierDef, reportStratifier, populations, groupDef);
+            valueOrNonSubjectValueStratifier(bc, stratifierDef, reportStratifier, populations, groupDef);
         }
     }
 
-    private static void nonCriteriaStratifier(
+    private static void valueOrNonSubjectValueStratifier(
             R4MeasureReportBuilderContext bc,
             StratifierDef stratifierDef,
             MeasureReportGroupStratifierComponent reportStratifier,
--- a/cqf-fhir-cr/src/test/java/org/opencds/cqf/fhir/cr/measure/r4/MeasureStratifierTest.java
+++ b/cqf-fhir-cr/src/test/java/org/opencds/cqf/fhir/cr/measure/r4/MeasureStratifierTest.java
@@ -8,6 +8,7 @@ import ca.uhn.fhir.rest.server.exceptions.InvalidRequestException;
 import org.hl7.fhir.r4.model.CodeableConcept;
 import org.hl7.fhir.r4.model.MeasureReport.MeasureReportStatus;
 import org.junit.jupiter.api.Test;
+import org.opencds.cqf.fhir.cr.measure.common.MeasurePopulationType;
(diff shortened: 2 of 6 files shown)


cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4PopulationBasisValidator.java
```diff
@@ -172,6 +171,15 @@ private void validateExpressionResultType(
         }
     }
 
+    private boolean doesBasisMatchResource(Class<?> resultClass, String groupPopulationBasisCode) {
+        // If we don't do this we'll fail with "boolean" vs. "Boolean"
+        if (resultClass == Boolean.class && BOOLEAN_BASIS.equals(groupPopulationBasisCode)) {
```
TipzCM: just a question, but does "equalsIgnoreCase" not work?

What about Integer vs int?

cqf-fhir-cr/src/test/java/org/opencds/cqf/fhir/cr/measure/r4/MeasureStratifierTest.java
```diff
@@ -704,4 +730,283 @@ void cohortResourceValueStratNull() {
...
+                .up()
+                .up()
+                .stratumByText("in-progress")
+                .firstPopulation()
+                .hasCount(1) // Patient-9 appears once in "in-progress"
+                .up()
+                .up()
+                .up()
+                .up()
+                .report();
```
TipzCM: This looks like common patterns here, but i have no idea what this test is doing.

Particularly all those 'ups'. Presumably it's going up to find a specific parent type, but i guess you just have to know that.

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4StratifierBuilder.java
```diff
@@ -67,14 +67,14 @@ private static void buildMultipleStratum(
...
+        if (stratifierDef.isCriteriaStratifier()) {
+            criteriaStratifier(bc, stratifierDef, reportStratifier, populations, groupDef);
         } else {
-            nonComponentStratifier(bc, stratifierDef, reportStratifier, populations, groupDef);
+            nonCriteriaStratifier(bc, stratifierDef, reportStratifier, populations, groupDef);
         }
     }
 
-    private static void componentStratifier(
+    private static void nonCriteriaStratifier(
```
Capt-Mac: Question: is nonCriteriaStratifier the same thing as valueStratifier?

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4PopulationBasisValidator.java
lukedegruchy: I want to make this explicit.  If this is a resource stratifier, I don't want "encounter" to match "Encounter."

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/r4/R4StratifierBuilder.java
lukedegruchy: It's a non-CRITERIA stratifier.  So it's either VALUE or NON_SUBJECT_VALUE.

Capt-Mac (approved):

PR #903, opened 2026-02-02
Fix Library/$evaluate to pass in canonical

barhodes:
We now  pass in the full canonical (including the version if present) of the Library resource rather than just the url when evaluating a Library.  If a version exists on the resource but doesn't get passed in to the engine, the Library that gets added to the cache has a null version.  This causes Library to not get cleared from the cache when the Library resource changes because the cache clearing method uses version in the lookup key.


lukedegruchy (approved):
One suggested small fix.

PR #910, opened 2026-02-05
Implement CRMI $create-changelog operation (R4)

Chris0296:
Added $create-changelog operation processors
Added operation documentation
Added tests

Issue tracking operation migration process: https://github.com/DBCG/aphl-vsm/issues/604

### DIFF
--- a/build-logic/src/main/kotlin/cqf.java-conventions.gradle.kts
+++ b/build-logic/src/main/kotlin/cqf.java-conventions.gradle.kts
@@ -16,6 +16,9 @@ java {
 }
 
 repositories {
+    if (file("${rootProject.projectDir}/local.properties").exists()) {
+        mavenLocal()
+    }
     mavenCentral()
     maven {
         name = "central-snapshots"


cqf-fhir-cr-hapi/src/main/java/org/opencds/cqf/fhir/cr/hapi/common/HapiCreateChangelogProcessor.java
```diff
@@ -0,0 +1,285 @@
...
+        this.packageProcessor = new PackageProcessor(repository);
+        this.hapiArtifactDiffProcessor = new HapiArtifactDiffProcessor(repository);
+    }
+
+    @Override
+    public IBaseResource createChangelog(
+            IBaseResource source, IBaseResource target, IBaseResource terminologyEndpoint) {
+
+        // 1) Use package to get a pair of bundles
+        ExecutorService service = Executors.newCachedThreadPool();
```
c-schuler: Creating a thread pool per request is expensive. Consider using a shared executor or virtual threads.

cqf-fhir-cr-hapi/src/main/java/org/opencds/cqf/fhir/cr/hapi/common/HapiCreateChangelogProcessor.java
```diff
@@ -0,0 +1,285 @@
...
+        try {
+            packages = service.invokeAll(Arrays.asList(
+                    () -> packageProcessor.packageResource(source, params),
+                    () -> packageProcessor.packageResource(target, params)));
+            sourceBundle = (Bundle) packages.get(0).get();
+            targetBundle = (Bundle) packages.get(1).get();
+            service.shutdownNow();
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new UnprocessableEntityException(e.getMessage());
```
c-schuler: Hmm, this exception would imply the client provided bad input, right? Maybe InternalErrorException would be more appropriate here (arguably for the ExecutionException below as well).

cqf-fhir-cr-hapi/src/main/java/org/opencds/cqf/fhir/cr/hapi/common/HapiCreateChangelogProcessor.java
```diff
@@ -0,0 +1,285 @@
...
+        Bundle targetBundle;
+        Parameters params = new Parameters();
+        params.addParameter().setName("terminologyEndpoint").setResource((Resource) terminologyEndpoint);
+        try {
+            packages = service.invokeAll(Arrays.asList(
+                    () -> packageProcessor.packageResource(source, params),
+                    () -> packageProcessor.packageResource(target, params)));
+            sourceBundle = (Bundle) packages.get(0).get();
+            targetBundle = (Bundle) packages.get(1).get();
+            service.shutdownNow();
```
c-schuler: This seems redundant with the `shutdown()` in the finally block.

cqf-fhir-cr-hapi/src/main/java/org/opencds/cqf/fhir/cr/hapi/common/HapiCreateChangelogProcessor.java
```diff
@@ -0,0 +1,285 @@
...
+                            "Type must be provided when adding an operation to the ChangeLog"));
+            var newValue = getParameter(change, "value");
+            var path = getPathParameterNoBase(change);
+            var originalValue = getParameter(change, "previousValue").map(o -> (Object) o);
+            // try to extract the original value from the
+            // source object if not present in the Diff
+            // Parameters object
+            try {
+                if (originalValue.isEmpty() && !type.equals("insert") && sourceResource != null && path.isPresent()) {
+                    originalValue = Optional.of((new BeanWrapperImpl(sourceResource).getPropertyValue(path.get())));
```
c-schuler: The use of BeanWrapperImpl seems fragile here. FHIR resource paths don't always map cleanly to Java bean property names. I would suggest using the FhirTerser for navigating FHIR resource paths. Alternatively, the project already uses ModelResolver.resolvePath() elsewhere for the same purpose.

cqf-fhir-cr-hapi/src/main/java/org/opencds/cqf/fhir/cr/hapi/common/HapiCreateChangelogProcessor.java
```diff
@@ -0,0 +1,285 @@
...
+            try {
+                bin.setContent(mapper.writeValueAsString(changelog).getBytes(StandardCharsets.UTF_8));
+            } catch (JsonProcessingException e) {
+                throw new UnprocessableEntityException(e.getMessage());
+            }
+
+            return bin;
+        }
+
+        return null;
```
c-schuler: If source and/or target are not found we should probably throw a meaningful exception, right? Seems like a possible NPE downstream.

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/common/CreateChangelogProcessor.java
```diff
@@ -0,0 +1,1374 @@
...
+import org.hl7.fhir.r4.model.Library;
+import org.hl7.fhir.r4.model.MetadataResource;
+import org.hl7.fhir.r4.model.Parameters;
+import org.hl7.fhir.r4.model.Period;
+import org.hl7.fhir.r4.model.PlanDefinition;
+import org.hl7.fhir.r4.model.PrimitiveType;
+import org.hl7.fhir.r4.model.RelatedArtifact;
+import org.hl7.fhir.r4.model.UsageContext;
+import org.hl7.fhir.r4.model.ValueSet;
+import org.hl7.fhir.r4.model.ValueSet.ConceptSetComponent;
```
c-schuler: This, being R4-specific, should not be in the version-agnostic common package. Move to R4 package(s) or make use of adapters.

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/common/CreateChangelogProcessor.java
```diff
@@ -0,0 +1,1374 @@
+package org.opencds.cqf.fhir.cr.common;
```
c-schuler, line 1: This class is huge. I would suggest extracting the ChangeLog and its inner classes into their own classes and the ValueSet code/leaf mapping logic into a dedicated helper.

PR #911, opened 2026-02-05
Refactor pre-library evaluation processing to use new CQL APIs.

lukedegruchy:
## Summary

This MR refactors the measure evaluation pre-processing pipeline to eliminate direct CQL engine state manipulation (`state.setParameter()`, `state.setParameters()`, `state.init()`) and the deprecated `MeasureProcessorUtils` instance class. Instead, user-supplied CQL parameters and measurement period now flow through `EvaluationParams` (the clean CQL engine API), and measurement period validation/resolution is extracted into stateless static utilities shared by both R4 and DSTU3 processors.

1. **Replaced `MeasureProcessorUtils` (instance class, deleted) with `MeasureProcessorTimeUtils` (static utility class)**: The old class mixed stateless time utilities with stateful CQL engine manipulation. The new class is purely static and separates measurement period validation (`validateAndConvertMeasurementPeriod`, takes ELM `Library` directly) from the deprecated CQL-default resolution path (`resolveDefaultMeasurementPeriodWithLibraryStack`, marked `@Deprecated(forRemoval = true)` with detailed rationale).
2. **Parameters now flow through `EvaluationParams` instead of pre-setting on CQL state**: The `setArgParameters()` / `preLibraryEvaluationPeriodProcessing()` methods in `R4MeasureProcessor` and the equivalent block in `Dstu3MeasureProcessor` are removed. Both processors now build a `parametersMap` and pass it through `MeasureEvaluationResultHandler.getEvaluationResults()` into `LibraryEngine.getEvaluationResult()` as `rawParameters`, where `CqlEngine.evaluate()` handles propagation to all libraries.
3. **DSTU3 processor aligned with R4 pattern**: The DSTU3 processor previously had its own separate library init/parameter-setting code path. It now uses the same shared `resolveMeasurementPeriodIntoParameters()` utility as R4.
4. **`R4MeasureProcessor` constructor simplified**: The `MeasureProcessorUtils` parameter was removed from the constructor, simplifying all call sites (CLI, Spring config, collect data service, multi-measure service).

## Code Review Suggestions

- [ ] Verify that `resolveParameterMap()` returning `Map.of()` for null parameters (R4MeasureProcessor) is safe for all callers — it's immediately wrapped in `new HashMap<>()` by `evaluateMultiMeasuresWithCqlEngine` but other callers of `resolveParameterMap` should be checked
- [ ] The `resolveMeasurementPeriodIntoParameters()` method assumes `libraryIdentifiers.get(0)` is the primary library — confirm this holds for multi-library measure evaluation scenarios where libraries may be in a different order
- [ ] The deprecated `resolveDefaultMeasurementPeriodWithLibraryStack()` still uses `getEvaluationVisitor().visitParameterDef()` (deprecated CQL API) — verify this is acceptable as a temporary measure until the CQL engine exposes `resolveParameterDefault()`
- [ ] `getDefaultMeasurementPeriod()` reads from `context.getState().getParameters()` which depends on parameters having been set during `evaluate()` — confirm this is only called after evaluation has run (used for report building)
- [ ] The `MeasureEvalType.getEvalType()` static method was moved from `MeasureProcessorUtils` — confirm no external callers outside this repo depended on the instance method
- [ ] `LibraryInitHandler` error message now uses singular/plural ("library"/"libraries") based on `ids.size()` — verify no downstream assertions depend on the exact wording beyond the tests updated in this MR

### DIFF
--- a/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/common/FunctionEvaluationHandler.java
+++ b/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/common/FunctionEvaluationHandler.java
@@ -358,10 +358,27 @@ public class FunctionEvaluationHandler {
             throw new InternalErrorException("StratifierDef component expression is missing.");
         }
         var stratifierExpression = componentDef.expression();
+
+        // Check if this is a function expression
+        // Per issue #909: Non-subject value stratifiers can now use BOTH:
+        // - CQL functions (for resource-level stratification)
+        // - Scalar expressions (for subject-level stratification)
+        // If it's a scalar expression, skip function processing here;
+        // MeasureEvaluator.handleNonBooleanBasisComponent() will use the fallback path
+        if (!isExpressionFunctionRef(context, libraryIdentifier, stratifierExpression)) {
+            logger.debug(
+                    "Non-subject value stratifier expression '{}' is a scalar expression, "
+                            + "skipping function processing for measure: {}",
+                    stratifierExpression,
+                    measureUrl);
+            return;
+        }
+
+        // This message is kept for safety - should not be reached since we validated above
         final String exceptionMessageIfNotFunction =
                 """
                 Measure: '%s', Non-subject value stratifier expression '%s' must be a CQL function definition, but it is not.
-                For non-boolean population basis, stratifier component criteria expressions must be "
+                For non-boolean population basis, stratifier component criteria expressions must be
                 CQL functions that take a parameter matching the population basis type.
                 """
                         .formatted(measureUrl, stratifierExpression);
--- a/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/common/MeasureMultiSubjectEvaluator.java
+++ b/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/common/MeasureMultiSubjectEvaluator.java
@@ -14,6 +14,7 @@ import java.util.Map.Entry;
 import java.util.Set;
 import java.util.stream.Collector;
 import java.util.stream.Collectors;
+import org.apache.commons.collections4.CollectionUtils;
 import org.hl7.fhir.instance.model.api.IBaseResource;
 import org.opencds.cqf.fhir.cr.measure.MeasureStratifierType;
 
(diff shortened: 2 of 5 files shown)


cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/common/LibraryInitHandler.java
```diff
@@ -77,10 +77,11 @@ private static List<CompiledLibrary> getCompiledLibraries(List<VersionedIdentifi
                             .formatted(ids, allErrors));
 
         } catch (CqlIncludeException exception) {
+            var libraryIds = ids.stream().map(VersionedIdentifier::getId).toList();
+            var label = ids.size() == 1 ? "library" : "libraries";
             throw new IllegalStateException(
-                    "Unable to load CQL/ELM for libraries: %s. Verify that the Library resource is available in your environment and has CQL/ELM content embedded."
-                            .formatted(
-                                    ids.stream().map(VersionedIdentifier::getId).toList()),
+                    "Unable to load CQL/ELM for %s: %s. Verify that the Library resource is available in your environment and has CQL/ELM content embedded."
```
TipzCM: why don't we say "for %d %s" (ie, "unable to load... for 2 libraries")

am i the only one who finds "unable to load ... for libraries: 2" confusing?

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/common/MeasureEvalType.java
```diff
@@ -62,6 +63,17 @@ public static Optional<MeasureEvalType> fromCode(String code) {
         return Optional.ofNullable(evalType);
     }
 
+    public static MeasureEvalType getEvalType(MeasureEvalType evalType, String reportType, List<String> subjectIds) {
+        if (evalType == null) {
+            evalType = MeasureEvalType.fromCode(reportType)
```
TipzCM: internal mutation of a value isn't nice.

Can we just return a new value that has this type?

TipzCM (approved):

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/common/MeasureEvalType.java
lukedegruchy: Fixed.

PR #922, opened 2026-02-18
Refactor $cql operation processor to be version agnostic

barhodes:
Refactored the processor for the $cql operation to be version agnostic and aligned with the other CR processors.  Added support for the prefetchData parameter for both the $cql and Library/$evaluate operations.


cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/cql/evaluate/CqlEvaluationRequest.java
```diff
@@ -0,0 +1,161 @@
...
+            IBaseBundle data,
+            List<? extends IBaseBackboneElement> prefetchData,
+            String content,
+            LibraryEngine libraryEngine,
+            ModelResolver modelResolver) {
+        checkNotNull(libraryEngine, "expected non-null value for libraryEngine");
+        fhirVersion = libraryEngine.getRepository().fhirContext().getVersion().getVersion();
+        this.libraryEngine = libraryEngine;
+        this.subjectId = subjectId;
+        if (expression == null && content == null) {
```
c-schuler: Should we check for empty expression string? If expression is "", would it fail downstream?

PR #995, opened 2026-04-06
Fix CLI module to publish unclassified fat JAR as primary artifact

c-schuler:
Since the Spring Boot Gradle plugin was applied to cqf-fhir-cr-cli, the standard jar task was disabled by default and only classified JARs (-plain, -javadoc, -module) were published to Maven Central. Downstream projects depending on cqf-fhir-cr-cli:jar (without a classifier) got 404s from Maven Central, breaking builds that unpack the CLI fat JAR.


brynrhodes (approved):

JPercival:
Tested locally, doesn't quite produce the same artifact layouts. Reused the idea in #996

PR #1002, opened 2026-04-21
Ensure Leaf ValueSet versions are stripped during $draft

Chris0296:
Currently in VSM, when a new program is drafted - pinned ValueSet versions are persisted from the original program into the new draft. This is not desirable, new drafts should default to using the latest version of the ValueSet - until the user manually pins a version. 

This change ensures freshly drafted programs do not have ValueSet versions pinned.

Includes updated tests for each FHIR version, modified test data to satisfy the 'isGrouper' check & addition of versioned 'isGrouper' methods.

### DIFF
--- a/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/crmi/KnowledgeArtifactProcessor.java
+++ b/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/crmi/KnowledgeArtifactProcessor.java
@@ -167,6 +167,18 @@ public class KnowledgeArtifactProcessor {
                         .anyMatch(uc -> uc.hasCode() && uc.getCode().getCode().equals(TransformProperties.grouperType));
     }
 
+    public static boolean isGrouper(org.hl7.fhir.r5.model.MetadataResource resource) {
+        return resource.getResourceType() == org.hl7.fhir.r5.model.ResourceType.ValueSet
+                && resource.getUseContext().stream()
+                        .anyMatch(uc -> uc.hasCode() && uc.getCode().getCode().equals(TransformProperties.grouperType));
+    }
+
+    public static boolean isGrouper(org.hl7.fhir.dstu3.model.MetadataResource resource) {
+        return resource.getResourceType() == org.hl7.fhir.dstu3.model.ResourceType.ValueSet
+                && resource.getUseContext().stream()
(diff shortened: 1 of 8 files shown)


sliver007 (approved):

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/crmi/KnowledgeArtifactProcessor.java
```diff
@@ -167,6 +167,18 @@ public static boolean isGrouper(MetadataResource resource) {
...
+                && resource.getUseContext().stream()
+                        .anyMatch(uc -> uc.hasCode() && uc.getCode().getCode().equals(TransformProperties.grouperType));
+    }
+
+    public static boolean isGrouper(org.hl7.fhir.dstu3.model.MetadataResource resource) {
+        return resource.getResourceType() == org.hl7.fhir.dstu3.model.ResourceType.ValueSet
+                && resource.getUseContext().stream()
+                        .anyMatch(uc -> uc.hasCode() && uc.getCode().getCode().equals(TransformProperties.grouperType));
+    }
+
```
barhodes, line 181: Rather than having multiple versions of this method perhaps look at using a Resource or ValueSet adapter here.

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/visitor/DraftVisitor.java
```diff
@@ -123,6 +124,31 @@ private List<IDomainResource> createDraftsOfArtifactAndRelated(
...
+                            KnowledgeArtifactProcessor.isGrouper(
+                                    (org.hl7.fhir.dstu3.model.MetadataResource) newResource);
+                        case R4 ->
+                            KnowledgeArtifactProcessor.isGrouper((org.hl7.fhir.r4.model.MetadataResource) newResource);
+                        case R5 ->
+                            KnowledgeArtifactProcessor.isGrouper((org.hl7.fhir.r5.model.MetadataResource) newResource);
+                        default ->
+                            throw new UnprocessableEntityException("Unsupported FHIR version: "
+                                    + repository.fhirContext().getVersion().getVersion());
+                    };
```
barhodes, line 141: Same here, using adapters should let you avoid using a switch on the FHIR version.

PR #1007, opened 2026-04-23
Preserve existing releaseLabel when not provided in Release operation

sliver007:
### Description
This change fixes an issue in the `ReleaseVisitor` where an existing `releaseLabel` extension could be unintentionally cleared when the `releaseLabel` parameter is omitted from the release operation.

#### What changed
- Updated handling of the `releaseLabel` parameter:
  - Previously: defaulted to an empty string (`""`) when not provided.
  - Now: treated as optional and only applied when explicitly present.
- `updateReleaseLabel` is now invoked **only if** a `releaseLabel` value is provided.

#### Why this matters
- Prevents accidental overwriting/removal of an existing `releaseLabel`.
- Aligns behavior with expected partial-update semantics (i.e., absence of a parameter should not mutate existing state).
- Improves safety and predictability of release operations.

#### Tests
- Added new test cases across DSTU3, R4, and R5:
  - `release_releaseLabel_not_provided_should_preserve_existing`
- These verify that:
  - An existing `releaseLabel` remains unchanged when the parameter is omitted.

#### Impact
- Backward-compatible change.
- Only affects scenarios where `releaseLabel` is not supplied.
- No change to behavior when `releaseLabel` **is** provided.


Chris0296 (approved):
Nice find. Looks good!

PR #1008, opened 2026-04-23
Enhance local expansion of hybrid definitions

sliver007:


cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/visitor/ExpandHelper.java
```diff
@@ -112,38 +111,83 @@ public void expandValueSet(
                         e.getMessage());
             }
         }
-        // Else if the ValueSet has a simple compose then we will perform naive expansion.
-        if (valueSet.hasSimpleCompose()) {
-            valueSet.naiveExpand();
+
+        // Hybrid local expansion for ValueSets with explicit concepts and/or valueSet
+        // references
+        if (valueSet.hasCompose() && (valueSet.hasExplicitConcepts() || valueSet.hasValueSetReferences())) {
```
c-schuler: Hmm. What happens with include filters here? It looks like the filters will be silently ignored in the hybrid path. In the old hasSimpleCompose/hasGroupingCompose branches, both required `noneMatch(csc -> csc.hasFilter())`. So previously include filters would fall through to $expand and fail if a terminology server was not specified. I think that is better failure mode than an incorrect expansion.

Exclude filters also appear to be ignored. So codes that should be excluded could leak into the expansion.

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/visitor/ExpandHelper.java
```diff
@@ -112,38 +111,83 @@ public void expandValueSet(
...
                         valueSet.get().getClass(),
                         "$expand",
                         (IBaseParameters) expansionParameters.get(),
                         valueSet.get().getClass(),
                         headers);
-                valueSet = (IValueSetAdapter) IAdapterFactory.createAdapterForResource(vs);
-                // Validate that the expansion parameters reflect what we asked for
+
+                var expandedAdapter = (IValueSetAdapter) adapterFactory.createResource(vs);
+                valueSet.setExpansion(expandedAdapter.getExpansion());
```
c-schuler: Old code reassigned the local valueSet to the $expand response, so valueSet.getUrl() / valueSet.getVersion() came from the response. The new code calls only valueSet.setExpansion(expandedAdapter.getExpansion()). terminologyServerExpand still does if (!valueSet.hasVersion()) valueSet.setVersion(expandedValueSet.getVersion()). The fallback path now no longer does this. Intentional? If not, mirror the version handling from terminologyServerExpand.

c-schuler (changes requested):
Looks mostly good. Note the possible issues I commented.

I would suggest adding some additional testing for those issues:
1. test for the filter + explicit-concept or filter + valueSet-reference case, which would likely currently pass with incorrect results.
2. test for a compose with exclude + explicit concepts, which would likely also pass with incorrect results
3. test confirming the fallback $expand path still preserves version

c-schuler (approved):
Looks good pending test fixes. One nit is that ExpandHelper.groupExpand() is no longer used - dead code.

Nice work!

PR #1009, opened 2026-04-24
Ensure Authoritative Source Extension is added to Leaf ValueSet during $release

Chris0296:
Currently in VSM when a program is imported, the Authoritative Source extension is added to all leaf ValueSets. 

If that program is cloned/drafted & released, the versioned ValueSets that are retrieved during the release process do not have the Authoritative Source extension added.

This means that if we clone this newly released program, any attempt to retrieve available versions of it's leaf ValueSets in VSM fails, since this is dependent on the Authoritative Source extension. 

This change ensures that any retrieved leaf ValueSets during the $release process will have the Authoritative Source extension added.

### DIFF
--- a/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/visitor/ReleaseVisitor.java
+++ b/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/visitor/ReleaseVisitor.java
@@ -1,5 +1,7 @@
 package org.opencds.cqf.fhir.cr.visitor;
 
+import static org.opencds.cqf.fhir.utility.Resources.newBaseForVersion;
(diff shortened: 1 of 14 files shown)


cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/visitor/ReleaseVisitor.java
```diff
@@ -544,6 +546,25 @@ private Optional<IKnowledgeArtifactAdapter> tryFindLatestDependency(
         return maybeAdapter;
     }
 
+    private void ensureAuthoritativeSourcExtension(IKnowledgeArtifactAdapter adapter) {
```
sliver007: Type - missing "e" at the end of Source

sliver007 (approved):

barhodes (approved):

PR #1011, opened 2026-04-26
Clinical Reasoning Dev Server

JPercival:
* Stripped down version of HAPI that shims in an InMemoryRepository behind the REST API


lukedegruchy (approved):
Approved with the caveat that we need to fix the sonar warnings and add a README.

.github/workflows/release-image.yml
```diff
@@ -0,0 +1,81 @@
+name: Release Image
+on:
+  push:
+    tags:
+      - 'v*'
+
```
lukedegruchy: Please add a README on how to use this.

cqf-fhir-cr-dev-server/src/main/java/org/opencds/cqf/fhir/cr/dev/server/config/ServerProperties.java
```diff
@@ -0,0 +1,37 @@
...
+ * <pre>
+ * cqf:
+ *   server:
+ *     base-path: /fhir          # servlet mount path
+ *     fhir-version: R4          # R4 or DSTU3
+ * </pre>
+ */
+@ConfigurationProperties(prefix = "cqf.server")
+public class ServerProperties {
+
```
lukedegruchy, line 18: Please fix the sonar warning on this class and other classes.

barhodes (approved):

PR #1019, opened 2026-04-29
Reject empty-stratifier Measures with a typed exception hierarchy

lukedegruchy:
## Summary

Real-world Measure resources exported from MADIE sometimes ship with
stratifiers whose `criteria` element has only `language` set — no
`expression` and no `component[]`. This is technically valid base FHIR
R4 (`criteria.expression` is `0..1`) but un-evaluable. Today these
resources crash deep inside `PopulationBasisValidator` with an unhelpful
Kotlin `Parameter specified as non-null is null` NPE thrown from
`EvaluationResult.get(null)`. This branch detects the malformed shape
at MeasureDef build time and surfaces it through a new typed exception
that names the offending stratifier and measure URL.

1. Tighten the criteria-presence check in `R4MeasureDefBuilder` so a
   stratifier with `criteria` present but no `criteria.expression` (and
   no components) is treated as missing criteria, allowing the existing
   "neither criteria nor component" branch to fire.
2. Introduce `InvalidMeasureDefinitionException` (a plain
   `RuntimeException` subclass) and use it in place of HAPI's
   `InvalidRequestException` at the empty-stratifier branch — error
   message now identifies the offending stratifier id alongside the
   measure URL.
3. Add a defense-in-depth blank-expression guard in
   `PopulationBasisValidator.validateExpressionResultType` so the
   Kotlin-non-null `EvaluationResult.get(...)` is never reachable with
   a `null` from any future call path (e.g. component criteria with
   null expression).
4. Add a regression test `MeasureStratifierTest.emptyStratifier` backed
   by a minimal synthetic `EmptyStratifier` Measure fixture (boolean
   basis, IP/Denom/Num populations, two empty stratifiers) that
   reproduces the MADIE-export pathology in isolation.

## Code Review Suggestions

- [ ] **Exception type change is observable to callers.** The
  empty-stratifier branch previously threw HAPI's
  `ca.uhn.fhir.rest.server.exceptions.InvalidRequestException`, which
  the FHIR REST layer maps to HTTP 400. It now throws
  `InvalidMeasureDefinitionException extends RuntimeException`, which
  unmapped will surface as HTTP 500. Confirm this is the desired wire-
  level behaviour, or that an upstream REST exception mapper translates
  the new type. The sibling `InvalidStratifierExpressionTypeException`
  and `StratifierExpressionNotFoundException` in the same package
  follow this same pattern, so this is consistent — but the "both
  criteria and components" guard a few lines above in
  `R4MeasureDefBuilder` still throws `InvalidRequestException`, which
  is now an inconsistency worth a conscious decision.
- [ ] **`hasCriteria` semantics shift has a side-effect on a sibling
  branch.** The local `hasCriteria` flag in
  `R4MeasureDefBuilder.getStratifierType` no longer just mirrors HAPI's
  `MeasureGroupStratifierComponent.hasCriteria()`; it now also requires
  `getCriteria().hasExpression()`. The same flag feeds the "both
  criteria and components" guard a few lines below — meaning a
  stratifier that has a criteria element with *only* a language plus
  components will no longer be rejected by that guard. Confirm that's
  intentional (it's arguably more permissive in a way the user didn't
  ask for).
- [ ] **Defensive blank-expression guard masks future bugs.**
  `PopulationBasisValidator.validateExpressionResultType` now silently
  returns when the stratifier expression is blank. Build-time
  validation should already catch this for stratifier criteria, but
  the guard also protects component criteria — and it does so without
  any logging. Worth considering a `log.warn` so this defense doesn't
  hide an upstream bug if one ever produces a blank expression where
  the build-time validation didn't run.
- [ ] **Sibling exception classes were not folded into a hierarchy.**
  An earlier draft of this work introduced a `MeasureEvaluationException`
  base class with `ExpressionResultNotFoundException` re-parented onto
  it. That was rolled back; only `InvalidMeasureDefinitionException`
  survives, extending `RuntimeException` directly. If a typed measure-
  evaluation exception hierarchy is desired in the future, this is the
  natural starting point.

### Setup

1. Build: `./gradlew :cqf-fhir-cr:assemble`.
2. Have a way to invoke `$evaluate-measure` via the HAPI/CR REST path
   (or directly via `R4MeasureProcessor` if that is the integration
   surface) — the wire-level checks below aren't covered by the JUnit
   test, only by the live evaluation flow.
3. If you have a MADIE-exported Measure (e.g. CMS165FHIR with empty
   stratifiers) handy, use it; otherwise the included `EmptyStratifier`
   test fixture is the minimal repro.

### Test cases

- [ ] **Empty stratifier on a deployed Measure surfaces a clear error.**
  Deploy a Measure containing a stratifier whose `criteria` has only
  `language` set (no `expression`, no `component[]`) to a running
  CR-enabled FHIR server. POST `$evaluate-measure`. Verify the response
  is an `OperationOutcome` whose diagnostics name the offending
  stratifier id and the measure URL — not a generic 500/NPE. Confirm
  the response status code matches your environment's expectation for
  the new exception type (it will be HTTP 500 unless an exception
  mapper has been wired for `InvalidMeasureDefinitionException`).
- [ ] **Truly-empty stratifier (no `criteria` element, no
  `component[]`) is still rejected.** This branch was already reachable
  before the change; only the exception type has changed. Confirm the
  diagnostic still includes the stratifier id and measure URL.
- [ ] **Stratifier with `criteria.expression` only continues to
  evaluate.** Re-run a known-good Measure (e.g. any of the existing
  `MeasureStratifierTest` fixtures) end-to-end through
  `$evaluate-measure` and confirm `MeasureReport.group.stratifier`
  output is unchanged from pre-branch behaviour. The `hasCriteria`
  semantics tweak should be a no-op here.
- [ ] **Stratifier with `component[].criteria.expression` only
  continues to evaluate.** Same as above for component-style
  stratifiers (multi-component / non-subject value). Confirm strata
  population counts are unchanged.
- [ ] **Borderline: criteria-with-only-language *plus* components.**
  Construct (or find) a Measure whose stratifier has a `criteria`
  element with only `language` set, AND has populated `component[]`
  entries. With the new `hasCriteria` semantics the "both criteria and
  components" guard will no longer fire for this shape — confirm the
  resulting evaluation is correct (the components should drive
  stratification and the empty top-level criteria should be ignored).
- [ ] **MADIE-exported measures already in your test corpus.** If you
  have any other MADIE-exported measure JSON files in your test
  corpus, smoke-evaluate them and confirm none trip the tightened
  `hasCriteria` check unexpectedly.


cqf-fhir-cr/src/test/resources/org/opencds/cqf/fhir/cr/measure/r4/MeasureStratifierTest/input/cql/EmptyStratifier.cql
JPercival, line 1: issue: too much/too unfocused content for a unit test this specific. Create a Measure resource inline and run it through the builder.

JPercival (approved):

PR #1023, opened 2026-05-01
Implement CRMI $validate operation

Chris0296:
Added $validate operation processors
Added FhirValidatorRegistry, which stores version specific validators, configured with an Npm Loaded ValidationSupportChain. This prevents running the expensive validator configuration process for each incoming request.

Tests to follow

Issue tracking operation migration process: https://github.com/DBCG/aphl-vsm/issues/604


cqf-fhir-cr-hapi/src/main/java/org/opencds/cqf/fhir/cr/hapi/config/FhirValidatorRegistry.java
```diff
@@ -0,0 +1,106 @@
...
+public class FhirValidatorRegistry {
+
+    private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(FhirValidatorRegistry.class);
+    private final DaoRegistry daoRegistry;
+    private final ConcurrentHashMap<FhirVersionEnum, FhirValidator> validators = new ConcurrentHashMap<>();
+    private final List<String[]> packagesToLoad = new ArrayList<>();
+
+    public FhirValidatorRegistry(DaoRegistry daoRegistry) {
+        this.daoRegistry = daoRegistry;
+        packagesToLoad.add(new String[] {"hl7.fhir.us.ecr", "2.1.2"});
```
c-schuler: Hmm. This should be configurable or moved into an ecr-specific package instead of the generic cqf-fhir-cr-hapi module. Take the package list (or CrSettings) as a constructor / Spring property so consumers can supply their own.

cqf-fhir-cr-hapi/src/main/java/org/opencds/cqf/fhir/cr/hapi/config/FhirValidatorRegistry.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    private @NonNull FhirInstanceValidator getFhirInstanceValidator(
+            FhirContext ctx,
+            NpmPackageValidationSupport npm,
+            PrePopulatedValidationSupport prePopulatedValidationSupport) {
+        var vsmIG = new NpmPackageValidationSupport(ctx);
+        var chain = new ValidationSupportChain(
+                npm,
+                vsmIG,
```
c-schuler: npm and vsmIG aren't doing anything, right? They are created, but nothing is ever loaded into them... The loadPrePopulatedValidationSupport method is actually loading the eCR profiles. Maybe I am misreading this?

cqf-fhir-cr-hapi/src/test/java/org/opencds/cqf/fhir/cr/hapi/common/HapiValidateProcessorTest.java
```diff
@@ -0,0 +1,82 @@
...
+        var ctx = FhirContext.forR4();
+        hapiValidateProcessor = new HapiValidateProcessor(ctx, fhirValidatorRegistry);
+
+        Assertions.assertThrows(UnprocessableEntityException.class, () -> {
+            hapiValidateProcessor.validate(null, null, null);
+        });
+    }
+
+    @Test
+    void validate_verify_operation() {
```
c-schuler, line 71: Would like to see this test actually assert behavior - e.g. inspect the returned operation outcome. I would also like to see tests for the FhirValidatorRegistry.

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/bundle/BundleProcessor.java
```diff
@@ -0,0 +1,33 @@
...
+    public BundleProcessor(IRepository repository, IValidateProcessor validateProcessor) {
+        this.repository = requireNonNull(repository, "repository can not be null");
+        this.fhirVersion = this.repository.fhirContext().getVersion().getVersion();
+        this.validateProcessor = validateProcessor;
+    }
+
+    public IBaseOperationOutcome validate(IBaseBundle resource, String mode, String profile) {
+        var processor = validateProcessor != null
+                ? validateProcessor
+                : new ValidateProcessor(FhirContext.forVersion(fhirVersion));
```
c-schuler: Just use repository.fhirContext() instead of a fresh context per call.

c-schuler (changes requested):
I have some concerns with the collision with HAPI's built-in $validate operation. I have been discussing with the committer through other channels. Despite that, the code looks pretty good - just a few suggested changes.

PR #1028, opened 2026-05-08
Feature selective data requirements

brynrhodes:
Added Group cohort definition capabilities and Group/$evaluate support


brynrhodes:
I don't understand the failure on ubuntu, those tests pass locally and here is the resource it is looking for:

https://github.com/cqframework/clinical-reasoning/blob/feature-selective-data-requirements/cqf-fhir-cr/src/test/resources/org/opencds/cqf/fhir/cr/shared/r4/group-sr/input/resources/DefaultSearchParameter.json

And the repository creation pointing to that spot:

https://github.com/cqframework/clinical-reasoning/blob/feature-selective-data-requirements/cqf-fhir-cr/src/test/java/org/opencds/cqf/fhir/cr/group/GroupProcessorTests.java#L66

barhodes (approved):

PR #1039, opened 2026-05-28
Harden $create-changelog and improve artifact-diff fidelity

c-schuler:
## Summary

Makes the `$create-changelog` operation (and the shared `$artifact-diff` engine it builds on) robust against real-world manifests and produces a cleaner, less noisy diff. Two threads of work: a series of null-safety fixes that were each blocking the operation on progressively more complex inputs, and diff-quality improvements that remove spurious and redundant entries from the result.
  
### Robustness (NPE fixes)

Each surfaced after the prior was unblocked, all on the diff/changelog path:

  1. **`HapiArtifactDiffProcessor.extensionEquals`** - dereferenced `extension.getValue()` unconditionally, NPE-ing on complex extensions that carry only nested sub-extensions and no top-level value (e.g. the `package-source` extension emitted by `$data-requirements`). Now null-guards both sides and falls back to `Extension.equalsDeep`.
  
  2. **`Page.addInsertOperation` / `addDeleteOperation` / `addReplaceOperation`** - dereferenced `oldData`/`newData`, which are legitimately null for pages representing a resource that exists on only one side of the diff (insert-only or delete-only). Now each side is guarded independently.
  
  3. **Null parameter names** - HAPI's `Parameters.getParameter(String)` NPEs when the list contains a component with a null name (which arises when a malformed canonical parses to null and is stored back on a component). Added a null-safe `hasParameterNamed` helper and switched the remaining `.getName().equals(...)` / `.equalsIgnoreCase(...)` call sites in both processors to literal-first form.
  
  ### Diff fidelity

  4. **Set-match contained expansion parameters** - the contained `Parameters.parameter[]` array is order-independent, but `FhirPatch.diff` compares arrays positionally, so a reordered manifest produced a cascade of meaningless `replace` ops pairing unrelated entries by index. Extended the existing match-and-reorder machinery (already used for `relatedArtifact`, `compose.include`, `expansion.contains`) to expansion parameters, matching by name + version-stripped URL. A pure reorder now produces no ops; a genuine version bump produces a single replace; add/remove produce clean insert/delete.

  5. **Skip unchanged dependencies** - `handleSourceMatches` recursed into and re-fetched every matched dependency even when the source and target referenced the identical version-pinned canonical (the same immutable artifact). Now short-circuits when the full canonicals are equal — no wasted fetch, no spurious sub-diff entry.
  
  6. **Suppress futile retrieval-failure placeholders** - when a recursive sub-diff couldn't resolve a side (commonly external code systems like LOINC/SNOMED/ICD/CPT that are never stored as full resources), the processor baked a `"Target/Source could not be retrieved"` string into the diff payload. These were both futile (external systems can't be content-diffed) and redundant (any reference-level change is already reported as a `relatedArtifact` operation). Now logged at debug instead of polluting the output.

### DIFF
--- a/cqf-fhir-cr-hapi/src/main/java/org/opencds/cqf/fhir/cr/hapi/common/HapiArtifactDiffProcessor.java
+++ b/cqf-fhir-cr-hapi/src/main/java/org/opencds/cqf/fhir/cr/hapi/common/HapiArtifactDiffProcessor.java
@@ -31,6 +31,7 @@ import org.hl7.fhir.r4.model.MetadataResource;
 import org.hl7.fhir.r4.model.Parameters;
 import org.hl7.fhir.r4.model.Parameters.ParametersParameterComponent;
 import org.hl7.fhir.r4.model.PlanDefinition;
+import org.hl7.fhir.r4.model.Reference;
 import org.hl7.fhir.r4.model.RelatedArtifact;
 import org.hl7.fhir.r4.model.StringType;
 import org.hl7.fhir.r4.model.Type;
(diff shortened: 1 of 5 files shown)


Chris0296 (approved):
Great improvements - nice work Chris!

PR #1041, opened 2026-05-29
Ensure date basis merges do not de-duplicate dates in population counts

### LINKED ISSUE #1040: MeasureReport deduping duplicate results generated by the cql
## What happened?

MeasureReport is outputting a different result than what the cql is returning. The initial population, from a dqm, is returning a list of two identical dates but the MeasureReport is deduping that list. Causing the actual vs expected results to mismatch.

The expectation is that the MeasureReport would return a count of 2 and not a count of 1 for Initial Population.

## How can we reproduce it?
define "Initial Population": {@2025-07-03, @2025-07-03}

## Environment

- Version / commit: CR 4.7.0
- OS / runtime: Mac OS 26.5

lukedegruchy:
* This is a tactical fix as the upgrade to use CQL 5.0.0 will be large and invasive, and there is already a clinical-reasoning with this work under way, so this PR is careful not to derail that work with large changes.
* Comment out CqlType-specific code in HashSetForFhirResourcesAndCqlTypes that deals with CqlTypes, as this fixes the case of de-duplicated dates.
* Add testing for duplicate strings and dates and ensure those tests pass, as well as tests for boolean basis
* Add tests for integer basis, but leave them disabled.
* Add a PRP file with a plan to address other de-duplicate cases.

Closes https://github.com/cqframework/clinical-reasoning/issues/1040

### DIFF
--- a/PRPs/prp-population-basis-primitive-duplicate-counting.md
+++ b/PRPs/prp-population-basis-primitive-duplicate-counting.md
@@ -1,0 +1,232 @@
+# PRP: Primitive-basis intra-subject duplicate counting (holistic fix for CDO-714)
+
+## Metadata
+- **Title**: Basis-aware per-subject storage so non-boolean, non-FHIR-Resource population values preserve intra-subject duplicates
+- **Status**: Proposed (deferred — blocked on CQL 5.0 / cql1 ExpressionResult type changes)
+- **Priority**: Medium (correctness, narrow surface today)
+- **Estimated Effort**: 2–3 days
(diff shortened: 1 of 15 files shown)


PRPs/prp-population-basis-primitive-duplicate-counting.md
antvaset, line 1: Looks good. `/Users/luke/development/smile-digital-health/git/clintel-repos/..` may have been included in repo paths by mistake?

antvaset (approved):

PR #1047, opened 2026-06-08
Add Conditional Custom eCR FhirValidator Configuration

Chris0296:
Part of: https://github.com/DBCG/aphl-vsm/issues/604

Configuration of custom FhirValidator Bean which is conditionally loaded, based on the property `cr.ecr.validate.enabled`

Validator support chain loads prepopulated validation support via Npm, with the packages to load being determined by the CrSettings `validatorPackages` field.

This FhirValidator is marked `@Primary`, so when it is conditionally created - it becomes the FhirValidator in use for FHIRs $validate operation.

### DIFF
--- a/cqf-fhir-cr-hapi/src/main/java/org/opencds/cqf/fhir/cr/hapi/config/CrBaseConfig.java
+++ b/cqf-fhir-cr-hapi/src/main/java/org/opencds/cqf/fhir/cr/hapi/config/CrBaseConfig.java
@@ -16,7 +16,8 @@ public class CrBaseConfig {
             EvaluationSettings evaluationSettings, TerminologyServerClientSettings terminologyServerClientSettings) {
         return new CrSettings()
                 .withEvaluationSettings(evaluationSettings)
-                .withTerminologyServerClientSettings(terminologyServerClientSettings);
+                .withTerminologyServerClientSettings(terminologyServerClientSettings)
+                .withValidatorPackage(new String[] {"hl7.fhir.us.ecr", "2.1.2"});
     }
 
     @Bean
--- a/cqf-fhir-cr-hapi/src/main/java/org/opencds/cqf/fhir/cr/hapi/config/CrEcrValidateConfig.java
+++ b/cqf-fhir-cr-hapi/src/main/java/org/opencds/cqf/fhir/cr/hapi/config/CrEcrValidateConfig.java
@@ -1,0 +1,113 @@
+package org.opencds.cqf.fhir.cr.hapi.config;
+
+import ca.uhn.fhir.context.FhirContext;
+import ca.uhn.fhir.context.support.DefaultProfileValidationSupport;
+import ca.uhn.fhir.jpa.api.dao.DaoRegistry;
+import ca.uhn.fhir.jpa.validation.ValidatorResourceFetcher;
+import ca.uhn.fhir.parser.IParser;
+import ca.uhn.fhir.validation.FhirValidator;
+import java.io.IOException;
+import java.io.InputStream;
+import org.hl7.fhir.common.hapi.validation.support.CommonCodeSystemsTerminologyService;
+import org.hl7.fhir.common.hapi.validation.support.InMemoryTerminologyServerValidationSupport;
+import org.hl7.fhir.common.hapi.validation.support.PrePopulatedValidationSupport;
+import org.hl7.fhir.common.hapi.validation.support.SnapshotGeneratingValidationSupport;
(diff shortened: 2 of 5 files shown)


cqf-fhir-cr-hapi/src/main/java/org/opencds/cqf/fhir/cr/hapi/config/CrEcrValidateConfig.java
```diff
@@ -0,0 +1,113 @@
...
+        for (var npmPackage : npmRepository.getLoadedPackages()) {
+            try {
+                var files = npmPackage.listResources("StructureDefinition", "ValueSet", "CodeSystem");
+                for (var filename : files) {
+                    logger.info("Loading pre-populated validation support from {}", filename);
+                    addResource(npmPackage, filename, parser, prePopulatedValidationSupport);
+                }
+
+            } catch (IOException e) {
+                logger.error("Error listing Resources from package {}", npmPackage.id(), e);
```
c-schuler, line 95: NpmRepository.ensurePackagesLoaded() catches all exceptions and leaves loadedPackages as an empty list. Then loadPrePopulatedValidationSupport catches IOException per package and continues. Then addResource catches all exceptions per resource (line 108). 

So if hl7.fhir.us.ecr#2.1.2 isn't in the local ~/.fhir cache and can't be downloaded, getLoadedPackages() returns empty, prePopulatedValidationSupport is empty. However, the @Primary FhirValidator is still created and installed, replacing the default. Combined with setValidateAgainstStandardSchema(false) / setValidateAgainstStandardSchematron(false), $validate then runs with effectively no profile support and silently stops doing meaningful validation for the lifetime of the server, with only a warning in the log. For a feature whose entire purpose is validation, "fail open and quiet" is probably the wrong default. Consider failing bean creation (or at least surfacing a hard error) when the configured packages don't load.

cqf-fhir-cr-hapi/src/main/java/org/opencds/cqf/fhir/cr/hapi/config/CrEcrValidateConfig.java
```diff
@@ -0,0 +1,113 @@
...
+     *
+     * <p>Marked {@link Primary} so it takes precedence when multiple {@link FhirInstanceValidator} beans
+     * are present in the application context.
+     */
+    @Bean
+    @Primary
+    public FhirInstanceValidator crEcrFhirInstanceValidator(
+            FhirContext fhirContext, DaoRegistry daoRegistry, CrSettings crSettings) {
+        NpmRepository npmRepository = new NpmRepository(fhirContext, crSettings.getValidatorPackages());
+        PrePopulatedValidationSupport prePopulatedValidationSupport = loadPrePopulatedValidationSupport(npmRepository);
```
c-schuler, line 53: This call chain: loadPrePopulatedValidationSupport then npmRepository.getLoadedPackages() at bean-construction time is problematic. FilesystemPackageCacheManager.loadPackage will attempt a network fetch from the package server when the package isn't cached, and then every StructureDefinition/ValueSet/CodeSystem in every transitive package is parsed into memory (all synchronously on the bean-init thread). This blocks application startup and can hang on a slow/unavailable package server. Worth at least documenting the pre-warm requirement, and ideally moving the load off the startup critical path.

cqf-fhir-cr-hapi/src/main/java/org/opencds/cqf/fhir/cr/hapi/config/CrBaseConfig.java
```diff
@@ -16,7 +16,8 @@ CrSettings settings(
             EvaluationSettings evaluationSettings, TerminologyServerClientSettings terminologyServerClientSettings) {
         return new CrSettings()
                 .withEvaluationSettings(evaluationSettings)
-                .withTerminologyServerClientSettings(terminologyServerClientSettings);
+                .withTerminologyServerClientSettings(terminologyServerClientSettings)
+                .withValidatorPackage(new String[] {"hl7.fhir.us.ecr", "2.1.2"});
```
c-schuler, line 20: The PR says packages "are determined by the CrSettings validatorPackages field," but the field is populated with a hardcoded id/version, not from any property. The only knob is the on/off flag, not which packages.

Just flagging this for now. I understand this will likely be addressed once the PR is ready for a final review.

c-schuler (changes requested):
Looking good so far. A few suggestions and we definitely need more testing. Nice work!

PR #1060, opened 2026-07-03
Add async processing capability to CMRI $package operation

Chris0296:
Adds optional asynchronous handling to the CRMI $package operation so clients can decouple request from response and avoid front-end timeouts on long-running packages. 

When a request includes `Prefer: respond-async`, the operation returns 202 Accepted immediately with a Content-Location header and runs the packaging work on a background thread; the client polls that location for the result. Without the header, behaviour is unchanged (synchronous).

This follows the general [FHIR Async Request Pattern](https://hl7.org/fhir/R4/async.html) handshake. Note the CRMI spec does not define async for $package and its return type is a single Bundle, so completion returns the Bundle inline (a 200) rather than the bulk-data manifest/file-download model, which does not apply here.

Changes
- PackageJob — job state, published as one immutable Snapshot via an AtomicReference for consistent, thread-safe reads.
- PackageJobService — in-memory job registry + bounded worker pool, with retention-based purging.
- AsyncPackageOperationHelper — detects Prefer: respond-async; runs sync or schedules async work, writing the response via HAPI's own RestfulServerUtils.streamResponseAsResource so content negotiation is identical to before.
- PackageJobStatusProvider — new $package-status system operation (202 while running, 200 + Bundle when done, 500 + OperationOutcome on failure, 404 for unknown jobs).
- Providers: all $package providers converted to @Operation(manualResponse = true) and delegated through the helper. Backward-compatible constructors added, so existing (non-async) wiring keeps working.
- Wiring: registered the job service, helper, and status provider beans in the R4/DSTU3 PackageOperationConfigs and the dev-server config.
- Off-thread safety: background work runs against a detached SystemRequestDetails (headers snapshotted) so it's safe to use after the servlet request is recycled.

Tests
- Integration test (PackageAsyncProviderIT): async kick-off returns 202 + Content-Location; polling yields 200 + Bundle; unknown job returns 404; existing synchronous PackageProviderIT confirms no regression.
- Unit tests for PackageJob, PackageJobService, and AsyncPackageOperationHelper.

Notes / limitations
- The job store is in-memory: jobs don't survive a restart. This has been deemed suitable for this implementation.


cqf-fhir-cr-hapi/src/main/java/org/opencds/cqf/fhir/cr/hapi/common/AsyncPackageOperationHelper.java
```diff
@@ -0,0 +1,131 @@
...
+     * Execute (or asynchronously schedule) a package operation and write the response.
+     *
+     * @param requestDetails the live request
+     * @param work performs the package operation against the supplied {@link RequestDetails}. For a
+     *     synchronous request this is the live request; for an async request it is a detached copy
+     *     safe to use after the originating request has returned.
+     */
+    public void packageOrRespondAsync(RequestDetails requestDetails, Function<RequestDetails, IBaseBundle> work) {
+        if (isAsyncRequested(requestDetails)) {
+            var jobId = jobService.submit(() -> work.apply(detach(requestDetails)));
```
c-schuler: `detach()` shouldn't be called inside the submitted supplier. 

Suggested fix:
```
var detached = detach(requestDetails);
var jobId = jobService.submit(() -> work.apply(detached));
```

cqf-fhir-cr-hapi/src/main/java/org/opencds/cqf/fhir/cr/hapi/common/AsyncPackageOperationHelper.java
```diff
@@ -0,0 +1,131 @@
...
+            throw new InternalErrorException("Failed to write $package response: " + e.getMessage(), e);
+        }
+    }
+
+    /**
+     * Produce a {@link RequestDetails} that is safe to use on a worker thread after the originating
+     * request has completed.
+     */
+    private static RequestDetails detach(RequestDetails requestDetails) {
+        var detached = new SystemRequestDetails(requestDetails);
```
c-schuler, line 129: Async path converts to SystemRequestDetails, which can bypass authorization/consent enforcement. 

SystemRequestDetails is treated as a trusted/internal request that may skip auth/consent enforcement. Therefore, a client that would be denied synchronously could get its package built by adding Prefer: respond-async. I don't see a test that asserts the async path preserves authorization.

We likely should align with the synchronous path and use ServletRequestDetails.

cqf-fhir-cr-hapi/src/main/java/org/opencds/cqf/fhir/cr/hapi/common/PackageJobService.java
```diff
@@ -0,0 +1,105 @@
...
+
+    public PackageJobService(int threadCount, Duration retention) {
+        this.retention = retention;
+        this.executor = Executors.newFixedThreadPool(threadCount, runnable -> {
+            var thread = new Thread(runnable, "package-async-worker");
+            // Daemon threads so a forgotten job never blocks JVM shutdown.
+            thread.setDaemon(true);
+            return thread;
+        });
+        Runtime.getRuntime().addShutdownHook(new Thread(this::shutdown, "package-async-shutdown"));
```
c-schuler: The service registers a JVM shutdown hook in its constructor but has no @PreDestroy/DisposableBean, never calls removeShutdownHook.shutdown(), and only runs at JVM exit. Therefore, on Spring context close, redeploy, etc leaks a 10-thread pool and adds a hook that pins the service instance for the life of the JVM.

Add:
`@PreDestroy void close() { shutdown(); }` and register the hook so it can be removed (or drop the hook entirely and rely on @PreDestroy).

c-schuler (changes requested):
This is looking good. I do have a couple change requests listed below. 

Additionally, it does not appear that ImplementationGuidePackageProvider (r4) was converted and silently ignores Prefer: respond-async. Perhaps that is still TODO?

cqf-fhir-cr-hapi/src/main/java/org/opencds/cqf/fhir/cr/hapi/r4/implementationguide/ImplementationGuidePackageProvider.java
```diff
@@ -22,48 +21,71 @@
...
-                .packageImplementationGuide(
-                        Eithers.forMiddle3(id),
-                        packageParameters(
-                                fhirVersion,
-                                getEndpoint(fhirVersion, terminologyEndpoint),
-                                usePut == null ? Boolean.FALSE : usePut.booleanValue()));
+        asyncHelper.packageOrRespondAsync(
+                requestDetails,
+                rd -> implementationGuideProcessorFactory
+                        .create(requestDetails)
```
c-schuler: This is a regression - using the outer requestDetails instead of the lamda's.

Should be `.create(rd)`

cqf-fhir-cr-hapi/src/main/java/org/opencds/cqf/fhir/cr/hapi/r4/implementationguide/ImplementationGuidePackageProvider.java
```diff
@@ -83,16 +109,18 @@ public IBaseBundle packageImplementationGuide(
...
-                                getIdType(fhirVersion, "Library", id),
-                                null),
-                        packageParameters(
-                                fhirVersion,
-                                getEndpoint(fhirVersion, terminologyEndpoint),
-                                usePut == null ? Boolean.FALSE : usePut.booleanValue()));
+        asyncHelper.packageOrRespondAsync(
+                requestDetails,
+                rd -> implementationGuideProcessorFactory
+                        .create(requestDetails)
```
c-schuler: This is a regression - using the outer requestDetails instead of the lamda's.

Should be `.create(rd)`

c-schuler (changes requested):
Looking better. A couple of regressions noted in the change requests

c-schuler (approved):

PR #1069, opened 2026-08-07
Prevent $release from accumulating duplicate depends-on entries on components

Chris0296:
Every $release appended a depends-on entry to a component artifact for each of its composed-of components, without checking whether one already existed. Because components are never stripped or deduped, the list grew on every release.

For eRSD: the rctc Library owns 6 grouper ValueSets as composed-of. Releasing took it from 6 composed-of + 6 depends-on to 6 composed-of + 12 depends-on. The next release would have produced 18, and so on.

Another issue is that the original had their resource version bumped, while their display text still was unchanged leaving half the entries with a display contradicting their reference.

This change removes any existing depends-on entries for the same component before adding the freshly built one (matching on version stripped URL + type).

Added test case for each ReleaseVisitor to validate behaviour.


barhodes (approved):

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/visitor/ReleaseVisitor.java
```diff
@@ -360,6 +360,12 @@ private void gatherDependencies(
 
                 // add to dependencies
                 var updatedRelatedArtifacts = artifactAdapter.getRelatedArtifact();
+                var newReference = IKnowledgeArtifactAdapter.getRelatedArtifactReference(componentToDependency);
+                updatedRelatedArtifacts.removeIf(ra -> Constants.RELATEDARTIFACT_TYPE_DEPENDSON.equalsIgnoreCase(
+                                IKnowledgeArtifactAdapter.getRelatedArtifactType(ra))
+                        && Objects.equals(
+                                Canonicals.getUrl(IKnowledgeArtifactAdapter.getRelatedArtifactReference(ra)),
+                                Canonicals.getUrl(newReference)));
```
c-schuler: These 2 calls to `Canonicals.getUrl()` can result in an NPE. Should probably be a bit more defensive here. It is an edge case, but represents a genuine regression

c-schuler (commented):
Looks good. Small note below

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/visitor/ReleaseVisitor.java
Chris0296: Nice catch Chris, thanks. I've updated the logic to be more defensive.

PR #1075, opened 2026-08-11
Fix $package falsely reporting resolved references as "not found in repository"

### LINKED ISSUE #1072: False "not found in repository" for a ValueSet that is in the package
## What happened?

ValueSet is present in the parameters, in the dependencies, and in the output bundle (both definition and expansion), yet the operation reports it as not in the repository. The message even says the resource "will not be included in package" while it *is* included.

A ValueSet that resolves and is included in the package should not be reported as "not found in repository." Either the version match should succeed, or it should not simultaneously be included in the bundle and reported as excluded if the resource truly can't be matched.

## How can we reproduce it?

Minimal steps:
1. Load a manifest/root artifact that depends on `http://hl7.org/fhir/ValueSet/ucum-vitals-common|4.0.1` (an HL7 core R4 ValueSet whose resource version is not string-equal to 4.0.1).
2. Invoke $package on the root artifact.
3. Observe the output bundle contains the ValueSet (definition + expansion), while the returned OperationOutcome contains an error: "Requested version '4.0.1' … not found in repository. Resource will not be included in package."

## Environment

- Version / commit: 4.10.0

## Anything else?

Related: https://github.com/DBCG/aphl-vsm/issues/708

c-schuler:
## Problem

During `$package`, a dependency that **is** included in the output bundle could still be reported in the manifest `OperationOutcome` as:

> Requested version '<v>' for resource '<url>' not found in repository. Resource will not be included in package.

## Root cause

`BaseKnowledgeArtifactVisitor.findResourceMatchingVersion` conflated **resolution** with **error reporting**. When the version-pinned repository search returned no exact match, it added the error *and* returned `null`. But the caller treats a `null` as "not resolved yet" and then falls back to the terminology server (`tryGetValueSetsFromTxServer`), which resolves the resource and packages it. So any reference resolved via the tx-server fallback (or otherwise present in the package) was falsely flagged.

## Fix

Separate resolution from reporting in `BaseKnowledgeArtifactVisitor`:

- `findResourceMatchingVersion(bundle, canonical)` now returns the version-matched resource or `null`
- A new `addUnresolvedReferenceMessage(...)` records the "not found in repository" error **only** when a reference is unresolvable by **any** path.
- The caller now resolves in order: repository search, terminology-server fallback and emits the message only if **both** fail.
- Removed the dead `messagesWrapper == null` back-compat branch (the sole caller always passes a non-null wrapper).

A reference is reported "not found" only when neither the repository nor the terminology server can resolve it.

## Testing

- New test reproduces #1072 (red before, green after): a version-pinned ValueSet missing from the local repo but supplied by the tx-server fallback is packaged **and** not flagged.
- Existing still passes: the error and exclusion are preserved.
- Full `:cqf-fhir-cr:test` suite and `spotlessCheck` pass.

## Notes

This is one of two `$package` issues from DBCG/aphl-vsm#708.

Closes #1072


cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/visitor/BaseKnowledgeArtifactVisitor.java
```diff
@@ -299,33 +300,29 @@ private IDomainResource findResourceMatchingVersion(
...
+     * repository or a terminology server. Only versioned references produce a message; unversioned
+     * references resolve to the latest available and are never reported here.
+     */
+    private void addUnresolvedReferenceMessage(String canonical, IBaseOperationOutcome[] messagesWrapper) {
+        var requestedVersion = Canonicals.getVersion(canonical);
+        if (requestedVersion == null || messagesWrapper == null) {
+            return;
+        }
+        var errorMessage = String.format(
+                "Requested version '%s' for resource '%s' not found in repository. Resource will not be included in package.",
```
Chris0296: Nitpick, but should this message acknowledge that it couldn't be resolved via Repository or Terminology Server?

Chris0296 (approved):

PR #1077, opened 2026-08-12
Fix $package skipping the configured terminologyEndpoint

### LINKED ISSUE #1073: Fix $package ignoring the configured terminologyEndpoint when a ValueSet's authoritative source differs
## What happened?

During `$package`, a ValueSet is not expanded and the outcome reports "Cannot expand ValueSet without terminology server" even though a terminology endpoint (tx.fhir.org) is configured. With `http://hl7.org/fhir/ValueSet/allergyintolerance-clinical|4.0.1`: it is in the parameters and dependencies, but the bundle contains only the definition (no expansion), plus the misleading "cannot expand without terminology server" message.

When a terminologyEndpoint (or matching artifactEndpointConfiguration) is supplied, $package should use it to expand the ValueSet rather than skipping the configured endpoint and reporting that no terminology server is available.

## How can we reproduce it?

Minimal steps:
1. Configure a terminologyEndpoint pointing at `http://tx.fhir.org/r4`.
2. $package a root artifact that depends on `http://hl7.org/fhir/ValueSet/allergyintolerance-clinical|4.0.1` - a ValueSet that (a) carries an authoritative-source-url extension whose normalized value differs from the endpoint address, and (b) composes with a filter/exclude so it cannot be expanded locally.
3. Observe: the ValueSet appears in the bundle as definition-only (no expansion), and the outcome contains "Cannot expand ValueSet without terminology server."

## Environment

- Version / commit: 4.10.0

## Anything else?

Related: https://github.com/DBCG/aphl-vsm/issues/708

c-schuler:
## Problem

During `$package`, a ValueSet could fail to expand with:

> Cannot expand ValueSet without terminology server: \<url>

**even though a `terminologyEndpoint` was configured.** 

## Root cause

The endpoint plumbing was intact. The failure was a decision gate in `ExpandHelper.expandValueSet`. The supplied `terminologyEndpoint` was only used when the ValueSet's `valueset-authoritativeSource` extension was absent or its address matched the endpoint's:

```java
if (terminologyEndpoint.isPresent()
        && (authoritativeSourceUrl == null
                || authoritativeSourceUrl.equals(terminologyEndpoint.get().getAddress()))) { ... }
```

A terminology server legitimately expands ValueSets authored by other authorities (hl7.org, VSAC), so for those the endpoint was silently skipped. When the ValueSet also could not be expanded locally (it has a filter/exclude), the code dropped to a repository $expand that failed with "Cannot expand ValueSet without terminology server". The comparison was also
asymmetric (the authoritative-source side was normalized via getAddressBase, the endpoint side was raw), so even the same server could fail to match on http/https, trailing-slash, or resource-path differences.

## Fix

In ExpandHelper:

- Keep the authoritative-source gate. For a grouper authored on a different server it correctly prefers local composition over hitting the endpoint (covered by existing tests). Removing it would break that behavior.
- Add the endpoint as a fallback when local expansion is not possible. The final fallback now first tries the configured terminologyEndpoint when it wasn't already used above. So a filter/exclude ValueSet with a mismatched authoritative source is expanded via the endpoint instead of failing.
- Normalize both sides of the address comparison with getAddressBase(...).

The change is scoped: it only alters behavior for the "not-locally-expandable + mismatched-authoritative-source + endpoint-present" case.

## Testing

- New test reproduces #1073 (red before, green after).
- New test locks in the CRMI path's independence from the extension.
- All existing ExpandHelperTest cases still pass.
- Full :cqf-fhir-cr:test suite and spotlessCheck pass.

Closes #1073


cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/visitor/ExpandHelper.java
```diff
@@ -177,8 +185,23 @@ public void expandValueSet(
...
-        // Fallback: repository $expand (for remaining ValueSets with compose)
+        // Fallback for ValueSets that could not be expanded locally.
         if (valueSet.hasCompose()) {
+            // A terminologyEndpoint was supplied but not used above because the ValueSet's authoritative
+            // source differs from it. Since local expansion was not possible, use the configured endpoint
+            // rather than failing — a general terminology server can expand ValueSets authored elsewhere.
+            if (terminologyEndpoint.isPresent() && !endpointIsAuthoritativeSource) {
+                try {
+                    terminologyServerExpand(valueSet, expansionParameters, terminologyEndpoint.get());
+                    expandedList.add(valueSet.getUrl());
```
Chris0296, line 202: Thoughts on using `addExpansionWarningParameter` here to explain the expansion was completed non-authoritatively?

Chris0296 (approved):

PR #1082, opened 2026-08-13
Fix empty/malformed component stratifiers for list-returning CQL functions

### LINKED ISSUE #1081: Component stratifier with a list-returning CQL function produces empty/malformed MeasureReport strata
## What happened?

A value-based (component) stratifier whose component criteria is a **CQL function that returns a list** produces malformed or empty strata in the MeasureReport. Reported against a HEDIS 2025 measure (NCQA `ENP-Reporting`) whose `Enrollment` stratifier has two components:

- `Age Stratifier(enrollmentDate Date)` returns a **scalar** (age category), and
- `Product Line Stratifier(enrollmentDate Date)` returns a **list**:

```sql
define function "Product Line Stratifier"(enrollmentDate Date):
  ("All reported member months info") mmInfo
    where mmInfo.monthlyEnrollmentDate = enrollmentDate
    return all mmInfo.payer.code        -- returns a List<code>
```

The list-valued component's stratum value is never unwrapped from the list. 

Per the CQM IG Conformance Requirement 3.17 (value-based stratifiers), each member is stratified by the tuple of component values, and the component value is the element value- not a raw CQL list representation. The report should contain per distinct tuple with clean component values, e.g.:

Age=18-19, ProductLine=MMO -> 2
Age=18-19, ProductLine=MCD -> 9
Age=18-19, ProductLine=MEP -> 1

A component function that returns a list should have defined semantics (single-element -> unwrap; empty -> a consistent absent/null stratum rather than a silent drop; -> fan out into multiple strata, as list stratifiers already do), and should never yield a silently empty stratifier.

## How can we reproduce it?

Minimal, self-contained (r4 measure IG folder):

- populationBasis = date, scoring = cohort, one initial-population returning of dates.
- One stratifier with two components:
  - Age Stratifier(Date) -> a scalar string (constant is fine),
  - Product Line Stratifier(Date) -> a list, e.g. `({ if (day from d) <= 2 then 'MMO' else if (day from d) <= 11 then 'MCD' else 'MEP' }) c return all c`.
- Evaluate for a single subject and assert the stratum component values are

Result: the ProductLine component value is 'MMO' (quoted) instead of MMO, an empty/real list values the stratifier collapses to no strata. A failing integration test of this shape is available.

## Environment

- Version / commit: reproduces on main (~4.10.1, CQL 5.1.0) and on v4.8.0

## Anything else?

- Scope: the defect is specific to NON_SUBJECT_VALUE (non-boolean basis) component stratifiers whose component expression is a CQL function returning a list.
- Requested fix on main, then backport to 4.8.1.
- Source: NCQA HEDIS 2025 ENP-Reporting.

c-schuler:
## Problem

A value-based (component) stratifier whose component criterion is a CQL **function that returns a list** produced corrupted stratum values. Reported against a HEDIS 2025 (NCQA `ENP-Reporting`) measure whose `Enrollment` stratifier pairs a scalar `Age Stratifier` with a list-returning `Product Line Stratifier`:

```sql
define function "Product Line Stratifier"(enrollmentDate Date):
  ("All reported member months info") mmInfo
    where mmInfo.monthlyEnrollmentDate = enrollmentDate
    return all mmInfo.payer.code        -- returns a List<code>
```

## Root cause

StratumValueWrapper normalizes a top-level CQL engine-native value at construction (CQL String -> plain String, FHIR ClassInstance -> HAPI type), but getValueAsString's Iterable branch recursed over the raw, un-normalized list elements. A list element left as a `org.opencds.cqf.cql.engine.runtime.String` matched no rendering branch and fell through to toString(), which renders with CQL quotes.

## Fix

Normalize at the top of `StratumValueWrapper.getValueAsString(Object)` so the recursion normalizes each list element too:

```java
private String getValueAsString(Object valueInner) {
    valueInner = normalizeEngineNativeValue(valueInner);
    ...
}
```

This handles every list flavor consistently: single-element -> the plain code; multi-element -> comma-joined codes; FHIR-code elements -> their code; empty/null already grouped via the existing empty/null sentinels. 

## Testing

- Unit regression 
- Integration reproduction - a date-basis component stratifier using the real ENP Product Line Stratifier
- Full :cqf-fhir-cr:test measure suite and spotlessCheck pass; no regressions.

Additionally validated (not included in this PR) end-to-end against the real NCQA ENP-Reporting measure + its full 12-library CQL closure + real patient Coverage data.

Closes #1081


barhodes (approved):

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/common/StratumValueWrapper.java
```diff
@@ -205,6 +205,11 @@ private String joinValues(String... elements) {
     }
 
     private String getValueAsString(Object valueInner) {
+        // Normalize engine-native values (CQL String -> plain String, FHIR ClassInstance -> HAPI type).
+        // The top-level value is normalized at construction, but list elements are not, so normalize
+        // here too: a list element left as a CQL String would otherwise fall through to toString() and
+        // render with CQL quotes (e.g. 'MMO' instead of MMO).
+        valueInner = normalizeEngineNativeValue(valueInner);
```
lukedegruchy, line 212: Nitpick:  I don't quite seem the same failure on your new StratifierMultiSubjectDateBasisTest when I comment out this line as Seena posted.  Instead of a completely empty stratifier, I see a stratifier with 3 strata and the test fails with a NullPointerException.

c-schuler: Yeah, you're correct. The updated code produced the correct output for the measure, but the issue wasn't faithfully reproduced. I added a test and further code enhancements to ensure the issue is properly addressed. Thanks!

c-schuler:
Follow-up: also fixed the literal empty stratifier.

Root cause of the empty stratifier: when a stratifier component's scalar-fallback expression resolves to null, handleNonBooleanBasisComponent logged a warning and returned without recording a result. Strata are built from each component's getResults() map, so a skipped component leaves no entry for that subject. When every component is null for every subject, buildSubjectResultsTable produces no rows resulting in zero strata and the stratifier serializes with no stratum array at all. This path is entirely upstream of StratumValueWrapper, which is why the value-rendering fix couldn't touch it.

lukedegruchy (approved):
I locally commented out your latest production code changes and then ran your latest test and saw this, which is what I expect.

    "stratifier": [ {
      "id": "enrollment-stratifier"
    } ]
  } ]

PR #1083, opened 2026-08-14
Ensure Changelog code entries carry a CodeCystem version

Chris0296:
## Problem

Every code in the changelog had an empty `version`, so the UI and spreadsheet showed a blank
CodeSystem Version column.

`updateCodeMapAndLeafMetadataMap` reads both `compose.include` and `expansion.contains`, but reads compose
first. In practice, with eRSD, the compose path never carries a CodeSystem version and the expansion path, which does carry the version, was then skipped by its `!codeMap.containsKey(code)` guard.

## Solution

`collectCodeSystemVersionsFromExpansion` builds a code -> version map from the ValueSet's own
expansion, and the compose path falls back to it when `compose.include` has no version. An explicit
compose version still wins.

Ordering is left alone deliberately: running the expansion path first would let a grouper's stored
expansion claim codes before the recursion into leaves, losing the per leaf version attribution.

## Tests

`ChangeLogTest` - Ensures correct version when: fallback applies, explicit compose version is preserved, null when neither side has one.

### DIFF
--- a/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/crmi/changelog/ChangeLog.java
+++ b/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/crmi/changelog/ChangeLog.java
@@ -127,6 +127,10 @@ public class ChangeLog {
             ValueSet valueSet,
             ArtifactDiffProcessor.DiffCache cache,
             ValueSetChild.Leaf leafData) {
+        // compose.include carries no code system version in practice, so fallback to the versions the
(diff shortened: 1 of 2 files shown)


c-schuler (approved):
Looks good. One thing I would flag as a possible issue is that `collectCodeSystemVersionsFromExpansion` keys the map by code alone, and the lookup (`codeSystemVersions.get(conceptReference.getCode())`) also ignores system.

This could lead to issues with VSs whose expansion contains the same code string under two systems. 

Does that make sense?

PR #1096, opened 2026-09-02
Update to CQL 5.3.0

antvaset:
- Use CQL 5.3.0 (future release).
- The updated engine returns `evaluatedResources` as `Map<kotlin.String, Value>` (with resources keyed by their string IDs) instead of `Set<Value>` (see https://github.com/cqframework/clinical_quality_language/pull/1834). This structure is reused where applicable (e.g. in `MeasureEvaluator`) to avoid building hashes for the evaluated resources.
- Update docs and `local.properties.example` following CQL's [repo restructure](https://github.com/cqframework/clinical_quality_language/blob/main/docs/proposals/repo-restructure.md).

### DIFF
--- a/.claude/skills/linked-builds-pr/SKILL.md
+++ b/.claude/skills/linked-builds-pr/SKILL.md
@@ -36,7 +36,7 @@ Read `local.properties` in the project root. Each property key maps to a `Linked
 in the registry. For example:
 
 ```properties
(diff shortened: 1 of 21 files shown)


antvaset:
Superseded by https://github.com/cqframework/clinical-reasoning/pull/1100

PR #1098, opened 2026-09-04
Feat: Surface all evaluated expressions as supporting evidence on subject MeasureReports

martinyushko-smile:
Adds `MeasureEvaluationOptions.supportingEvidenceMode` (`OFF | DECLARED | ALL_EXPRESSIONS`, default `DECLARED`). Under `ALL_EXPRESSIONS`, every top-level expression that he CQL engine already evaluates is written as a report-level `cqf-supportingEvidence` extension on subject-mode MeasureReports. The existing feature is used, and extended. `DECLARED` is the default, and current behavior. Expressions declared via `cqf-supportingEvidenceDefinition` stay at population level and are not duplicated. Under `OFF`, no supporting evidence is generated anywhere.

The selection step and type filter live as version-neutral changes, while report writing and value encoding necessarily stay per-version. The value filter scopes out FHIR elements and the merged function-evaluation accumulators. FHIR resources are encoded as reference strings.

This also incidentally closes standing encoder gaps on the shared supporting-evidence path (these previously failed the whole $evaluate-measure operation when declared, and for all intents and purposes, were not supported):
- Quantity, Ratio, Concept, non-temporal Interval — now delegate to the CQL engine's `FhirTypeConverter`
- Time — now valid FHIR, `14:30:00`, where previously emitted as an illegal CQL literal (`@T14:30:00`)
- Long — `valueString`
- ValueSet/CodeSystem — `valueCanonical`, `|version`-suffixed when pinned
- Resource-valued leaves now read their reference off the `ClassInstance`

And lastly, Code serialization is unchanged. It keeps its existing string form for compatibility.

The great majority of the diff is tests + test resources to cover the new cases.

Testing: report-level behaviour per mode, one declared-path encoding test per previously-failing type, full `cqf-fhir-cr` suite green. Validated end-to-end on a 400-member HEDIS cohort (90,981 resources) against local MinIO: 88 evidence entries per subject report, all names resolving to library defines, +6% report size on this sample. Backwards-compatibility was also tested, and default/existing behaviour remains untouched.

### DIFF
--- a/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/MeasureEvaluationOptions.java
+++ b/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/MeasureEvaluationOptions.java
@@ -3,6 +3,7 @@ package org.opencds.cqf.fhir.cr.measure;
 import java.util.HashMap;
 import java.util.Map;
 import org.opencds.cqf.fhir.cql.EvaluationSettings;
+import org.opencds.cqf.fhir.cr.measure.common.SupportingEvidenceMode;
 import org.opencds.cqf.fhir.utility.ValidationProfile;
 
 public class MeasureEvaluationOptions {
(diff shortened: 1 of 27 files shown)


martinyushko-smile:
Superseded by #1099, which pushes the identical commit (c1f625e) to a branch on this repo instead of a fork.

The two failing checks here were fork-permission artifacts rather than code problems: `spotlessCheck` itself passed, and the job only went red on the `add-pr-comment` step because fork PRs get a read-only `GITHUB_TOKEN`. Likewise `compile-against-cdr-cr` failed with an empty `TRIGGER_TOKEN`, since secrets aren't injected into fork runs. Both need a same-repo head to pass.

PR #1101, opened 2026-09-07
fix: preserve Coding answers during profile-constrained extraction

rob-reynolds:
When a QuestionnaireResponse contains a Coding answer and its extraction profile constrains an Observation value to CodeableConcept, extraction currently drops the answer. This can leave a subsequent PlanDefinition apply paused despite a supplied answer.

Convert Coding to CodeableConcept before assignment when the assignment parent matches the resource profile and the selected element is unsliced and has exactly one type. Preserve the complete Coding and repeated-property cardinality. Nested Extension and sibling-type-slice regressions prevent borrowing an unrelated profile constraint. Existing unsupported conversions and their diagnostic behavior remain unchanged.

### Validation

- Reproduced the original failure on unmodified main and v4.7.0 baselines.
- 21 conversion tests across DSTU3, R4 and R5, including 27 real extraction-entry cases, all passing. Cases cover differential/snapshot profiles, both choice-path spellings, metadata, repeated values, Boolean false and unsupported slices/multi-type elements.
- Combined with the separate generated-ID correction, main CR + utility suites: 3,626 passed, 24 skipped, zero failures/errors. Final test-only assertion rerun: 21 passed.
- A pinned-4.7 runtime comparison exercised 12 apply calls across original, Coding-only and combined configurations. Corrected configurations pause on missing data, reach the expected affirmative/negative activity after answers, and pause again after clearing an answer. Complete returned values and issue sets were compared.
- Formatting and main-source Checkstyle passed.

This PR contains only the Coding change. Generated IDs are in a separate stacked PR. The tested v4.7.0 backport is available on codex/qr-extraction-backport; upstream currently has no 4.7 maintenance branch to target. No installed runtime or release was replaced.

### DIFF
--- a/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/questionnaireresponse/extract/ProcessDefinitionItem.java
+++ b/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/questionnaireresponse/extract/ProcessDefinitionItem.java
@@ -560,6 +560,20 @@ public class ProcessDefinitionItem {
             IBase answerValue,
             IStructureDefinitionAdapter profile) {
         try {
+            // A parent-relative path can use this resource profile only at its resource root.
+            // Nested/sliced profile resolution remains unchanged; do not borrow a root value constraint.
+            if (profile != null
+                    && profile.getType().equals(parent.get().fhirType())
+                    && "Coding".equals(answerValue.fhirType())) {
+                var element = profile.getElementByPath(answerPath.split(":")[0]);
+                if (element != null
+                        && element.getType().size() == 1
+                        && StringUtils.isBlank(element.getSliceName())
+                        && !StringUtils.contains(element.getId(), ":")
+                        && "CodeableConcept".equals(element.getTypeCode())) {
+                    answerValue = transformValueToResource(request.getFhirVersion(), answerValue);
+                }
+            }
             parent.setValue(answerPath, transformAnswer(request, pathDefinition, answerValue, answerPath, profile));
         } catch (Exception e) {
             if (pathDefinition instanceof RuntimeChildPrimitiveDatatypeDefinition definition
--- a/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/questionnaireresponse/extract/ProcessDefinitionItem.java
+++ b/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/questionnaireresponse/extract/ProcessDefinitionItem.java
@@ -569,7 +569,7 @@ public class ProcessDefinitionItem {
                 if (element != null
                         && element.getType().size() == 1
                         && StringUtils.isBlank(element.getSliceName())
-                        && !StringUtils.contains(element.getId(), ":")
+                        && !StringUtils.contains(element.getId(), ':')
                         && "CodeableConcept".equals(element.getTypeCode())) {
                     answerValue = transformValueToResource(request.getFhirVersion(), answerValue);
                 }
--- a/cqf-fhir-cr/src/test/java/org/opencds/cqf/fhir/cr/questionnaireresponse/extract/ProfileAnswerConversionTest.java
+++ b/cqf-fhir-cr/src/test/java/org/opencds/cqf/fhir/cr/questionnaireresponse/extract/ProfileAnswerConversionTest.java
@@ -53,6 +53,57 @@ class ProfileAnswerConversionTest {
         assign(version, "CodeableConcept", "Coding", CODING, true, "category");
     }
 
+    @ParameterizedTest
+    @EnumSource(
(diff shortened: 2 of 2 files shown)


JPercival (approved):

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/questionnaireresponse/extract/ProcessDefinitionItem.java
c-schuler, line 640: The root cause of the defect occurs on this line. Fortunately, the else branch here already routes through `transformValueToResource`, which is identity for non-Coding. Making the first branch do the same looks like it'd fix this at the root, cover nested value[x] targets, and let the new `setAnswerValue` guard drop out. Is there a reason you preferred the call-site guard?

Something like this should work (for reference):

```java
if (answerType != null && !answerValue.fhirType().equals(answerType)) {
    var converted = transformValueToResource(request.getFhirVersion(), answerValue);
    if (converted.fhirType().equals(answerType)) {
        answerValue = converted;                 // Coding->CodeableConcept: now matches
    } else {
        var newAnswerValue = createBase(newBaseForVersion(answerType, ...));
        newAnswerValue.setValue(VALUE_PATH, answerValue);   // unchanged fallback + diagnostics
        answerValue = newAnswerValue.get();
    }
}
```

barhodes (changes requested):
Would like to look at adding this into the definition based population branch to avoid merge conflicts down the road.

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/questionnaireresponse/extract/ProcessDefinitionItem.java
rob-reynolds: Thanks — I tested this exact change against current PR head `066ae897`, removing the `setAnswerValue` guard and putting the conversion in `transformAnswer` as suggested (retaining the original value for the fallback).

The unchanged implementation passes all 21 conversion tests; the proposal passes 15, with the same two failing vectors across DSTU3/R4/R5.

The reason for the call-site guard is target context: `getElementByPath("value[x]")` is relative to the resource profile. In `extensionSliceCannotBorrowRootValueConstraint`, the assignment parent is an Extension, but the profile is an Observation whose root `value[x]` is CodeableConcept. The proposed conversion then produces `Extension.valueCodeableConcept` using that unrelated root constraint. The referenced Extension profile has not been resolved to establish its value type. The parent-type check prevents that borrowed-type conversion.

This does not mean the existing nested lookup is correct: it still borrows the root type, then its old fallback attempts to assign a nonexistent CodeableConcept.value and leaves the Extension answer absent. The guard limits the newly successful conversion to a root assignment; it does not repair nested extraction. The proposed change turns that existing unresolved case into a populated value without resolving its actual profile.

The other failure is the multi-type CC-first characterization. I would not treat conversion into an allowed CodeableConcept as inherently wrong; the broader concern is choosing from the first profile type rather than resolving the target consistently. That outcome alone is not my reason for retaining the guard. I have clarified that test's assertion message to identify the current patch's scope limit rather than imply a conformance prohibition.

I agree `transformAnswer` is a natural place for conversion once the actual target type is reliably scoped. With its current inputs, however, merely checking the converted value against `answerType` also accepts the borrowed root type. I've retained the scoped implementation for this PR rather than widening nested/slice resolution here.

I also added an actual extraction regression with two distinct Coding answers into repeating Observation.category, comparing both values and their complete metadata across all three versions. With the current implementation the expanded conversion suite passes 24/24. No production behavior or installed engine changed in this follow-up.

rob-reynolds:
@barhodes I checked `feature-definition-based-population` at `336c4e14`. Its `ProcessDefinitionItem.java` matches this PR's implementation apart from the existing Coding conversion guard, and it does not yet contain `ProfileAnswerConversionTest`. A scoped patch containing that guard and the conversion tests applies cleanly to that branch (verified against a temporary Git index; I have not run that branch's full suite).

I'm happy for this fix to land through the definition-based population work. Do you prefer a focused cherry-pick/patch there, or retargeting this PR after the branch is brought up to date? I have left the PR base and your branch unchanged because #1102 and #1104 are stacked above this work.

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/questionnaireresponse/extract/ProcessDefinitionItem.java
c-schuler: Ok, thanks for running the suggestion and relaying the results! Agreed that the actual "root fix" is reliable target-type resolution for nested/extension/slice contexts, not just relocating the call. Can we create a ticket for that?

rob-reynolds:
Superseded by #1108, which targets feature-definition-based-population.

PR #1102, opened 2026-09-07
fix: bound generated questionnaire and extraction resource IDs

rob-reynolds:
QuestionnaireResponse and extraction IDs are composed from several individually valid inputs, so the final ID can exceed FHIR's 64-character limit. Composing a response ID from a qualified/versioned Questionnaire ID can also append the subject suffix to the version component instead of creating a new logical response identity.

Normalize complete generated IDs at their final creation sites. Preserve already-valid IDs; otherwise use a sanitized, bounded stem plus 12 hexadecimal SHA-256 characters derived from the complete original input. Generate QuestionnaireResponse IDs from the Questionnaire's logical ID. Supplied IDs, subject references, canonicals and linkIds remain unchanged; generated references read the resulting resource identity.

Targets `feature-definition-based-population`. The Observation ID correction uses that branch's unified extractor. This PR contains only the generated-ID correction and is independent of #1108's profile-answer conversion changes.

### Scope

QuestionnaireResponse, extracted resources, extraction Bundle and contained OperationOutcome IDs. The short digest is deterministic, not collision-proof. Existing id-less outcome suffix behavior is retained. This does not repair extraction transaction URLs or establish transaction persistence readiness.

### Validation

- On target branch base `e7c5179045d347407250312f33ec2171056e645c`: 28 affected tests passed, zero failures/errors/skips, covering population, definition/Observation extraction, and generated IDs.
- Includes 11 generated-ID/reference cases across supported FHIR versions and one helper boundary test, including actual qualified/versioned Questionnaire ID parsing.
- Java formatting, affected main-source Checkstyle, and diff whitespace checks passed.
- No installed CRL runtime or release was replaced by this PR update.

PR #1104, opened 2026-09-08
Pause ordered applicability on unknown by default

rob-reynolds:
An unresolved applicability condition currently excludes its descendants, but an ordered `any` group can skip it and select an unconditional fallback. In an interactive workflow this can produce a recommendation and gather later questions before the missing answer is supplied.

Add `CrSettings.pauseOnUnknownApplicability`, defaulting to **true**, and propagate it through ordinary action evaluation and nested PlanDefinition requests. Unknown keeps the reached action's input questions but excludes its descendants/definition and stops later ordered alternatives. Known false still permits the next alternative; independent `all` actions remain independent. Hosts can explicitly set the flag to false for the previous non-applicable-on-null behavior. This is a missing-data runtime policy in addition to the first-true selection specified by [FHIR-50150](https://jira.hl7.org/browse/FHIR-50150), not a claim that the resolution explicitly mandates null pausing.

Nullable conjunction preserves false versus unknown. Malformed, non-Boolean or multi-valued conditions and evaluation failures report OperationOutcome errors and block ordered fallthrough, including when another condition is false. Each reached condition is evaluated once.

Validation: all 70 PlanDefinition tests pass on this branch and the 4.7 backport; main checkstyle passes and Java formatting was applied (test checkstyle is disabled by the existing build). Twelve native R4 applyR5 controls verify default/explicit settings, true/false/null, and collection results containing null, using the compiled 4.7 patch classes with the pinned 4.7 runtime. An additional unchanged customer-policy fixture returns three reached question groups rather than nine at its first unresolved condition, with no recommendation. This is API-level evidence, not a client-rendering or full clinical revalidation claim.

Caller-supplied Questionnaire items are retained. In a clear-answer replay, trimming only the supplied Questionnaire to three reached groups while retaining the full QuestionnaireResponse and repository returned those three groups with no recommendation or errors. Retention of explicitly supplied items does not establish a separate engine defect; this patch prevents new traversal beyond an ordered pause.

Stacked on #1102 for the previously tested extraction/ID fixes. After that PR merges, retarget this PR to main. The corresponding 4.7 source is pushed as `codex/apply-pause-on-unknown-r4`; no release or installed-runtime update is included.

### DIFF
--- a/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/CrSettings.java
+++ b/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/CrSettings.java
@@ -6,6 +6,28 @@ import org.opencds.cqf.fhir.utility.client.TerminologyServerClientSettings;
 public class CrSettings {
     private EvaluationSettings evaluationSettings;
     private TerminologyServerClientSettings terminologyServerClientSettings;
+    private boolean pauseOnUnknownApplicability = true;
+
+    /**
+     * Whether ordered ANY action groups stop at an unresolved applicability condition.
+     * Defaults to true. When enabled, the unresolved action's
+     * inputs remain available, but its descendants and later alternatives are not applied.
+     * Disabling this setting treats unknown as non-applicable and permits ordered fallthrough.
+     * Independent ALL actions remain independently evaluated. This does not remove items
+     * from a Questionnaire supplied by the caller.
+     */
+    public boolean isPauseOnUnknownApplicability() {
+        return pauseOnUnknownApplicability;
+    }
+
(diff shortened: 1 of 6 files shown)


rob-reynolds:
@brynrhodes @barhodes @c-schuler Are we OK to default the applicability any to "pause" on null?  This matches the descending behavior so it seems like the correct thing to do for the same level behavior.  But I don't want to break anything.  Note, there is a setting.

barhodes (changes requested):
I would like to discuss this on the CPG Project call.

c-schuler:
I have concerns about this behavior being opt-out instead of opt-in.

rob-reynolds:
Superseded by discussion in CPG: https://jira.hl7.org/browse/FHIR-58945.

PR #1105, opened 2026-09-08
Key the SDE and function-handler evaluated resource sets by identity

c-schuler:
### Problem

Three collections that accumulate engine `Value`s were plain `HashSet`s, so every insert recursively hashed a `ClassInstance`'s entire element tree. For FHIR resources that is a walk of the whole resource graph, per resource, per subject.

- `SdeDef.allEvaluatedResources` - `accumulate()` merges every subject's evaluated resources into this one set
- `FunctionEvaluationHandler`, measure-observation accumulation
- `FunctionEvaluationHandler`, stratifier-function accumulation

#1100 keyed the other evaluated-resource collections by resource identity; these three were missed.

### What this does

Switches all three to `HashSetForFhirResourcesAndCqlTypes`, so membership is decided by the `IdentityKey` machinery #1100 introduced rather than by a deep structural hash. No new types and no behavioural change - the same values are retained, and the same duplicates collapse.

Also removes a duplicated `addAll` in the stratifier path, where `evaluated` and `functionResult.getEvaluatedResources()` are the same collection assigned three lines apart, so the set was populated twice. That is a no-op for correctness and pure waste on the hot path.

### Evidence

Found by JFR profiling a 400-member HEDIS `AAB-Details` population run: every one of the 567 deep-hash samples resolved to `SdeDef.accumulate`, via `MeasureEvaluationResultHandler.processResults` and `MeasureMultiSubjectEvaluator.postEvaluationMultiSubject`.

After the change, `StructuredValue.hashCode` accounts for 0.0% of samples, down from 15.4%. The measure-report build phase improves to a normalised ratio of 0.609 against an unpatched build measured in the same interleaved sweep.

This cost scales with how many evaluated resources reach report building, so it is most visible on measures that retrieve broadly - `AAB-Details` retrieves `[Claim]` unfiltered, and claims were 64% of the test dataset.

### Verification

Full `:cqf-fhir-cr:test` suite shows no regressions against a clean `main` baseline captured on the same commit: 412 failing with this change versus 415 without, the difference being three known-flaky tests. Correctness held across twelve end-to-end measure evaluations - 400 of 400 members, zero errors, populations unchanged.

### DIFF
--- a/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/common/FunctionEvaluationHandler.java
+++ b/cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/common/FunctionEvaluationHandler.java
@@ -273,7 +273,9 @@ public class FunctionEvaluationHandler {
         // FhirResourceAndCqlTypeUtils.areObjectsEqual where needed; nothing in the downstream pipeline
         // does random-access lookup by input, so a List is sufficient and self-documenting.
         final List<ObservationEntry> functionResults = new ArrayList<>();
-        final Set<Value> evaluatedResources = new HashSet<>();
+        // Identity-keyed: these accumulate engine values whose hashCode walks the whole
+        // element tree. See HashSetForFhirResourcesAndCqlTypes.
+        final Set<Value> evaluatedResources = new HashSetForFhirResourcesAndCqlTypes<>();
 
         final String exceptionMessageIfNotFunction = """
             Measure: '%s', MeasureObservation population expression '%s' must be a CQL function
@@ -407,7 +409,9 @@ public class FunctionEvaluationHandler {
             // this will be used in MeasureEvaluator (Criteria population Id and Stratifier Expression)
             var expressionName = popDef.id() + "-" + stratifierExpression;
             final List<FunctionResultEntry> functionResults = new ArrayList<>();
-            final Set<Value> evaluatedResources = new HashSet<>();
+            // Identity-keyed: these accumulate engine values whose hashCode walks the whole
+        // element tree. See HashSetForFhirResourcesAndCqlTypes.
+        final Set<Value> evaluatedResources = new HashSetForFhirResourcesAndCqlTypes<>();
 
             for (var result : resultsIter) {
                 final ExpressionResult functionResult = evaluateNonSubValueStratifiersFunction(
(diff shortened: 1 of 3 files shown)


cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/common/FunctionEvaluationHandler.java
```diff
@@ -426,7 +429,6 @@ private static void processNonSubValueStratifier(
                             + "' returned null evaluatedResources for measure: " + measureUrl);
                 }
                 evaluatedResources.addAll(evaluated);
-                evaluatedResources.addAll(functionResult.getEvaluatedResources());
             }
```
lukedegruchy, line 432: I ran all test tests and did a full clinical-reasoning build + test and this passed.  So this didn't break anything, but I don't understand why.

c-schuler: Because it is a no-op

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/measure/common/FunctionEvaluationHandler.java
```diff
@@ -407,7 +408,9 @@ private static void processNonSubValueStratifier(
             // this will be used in MeasureEvaluator (Criteria population Id and Stratifier Expression)
             var expressionName = popDef.id() + "-" + stratifierExpression;
             final List<FunctionResultEntry> functionResults = new ArrayList<>();
-            final Set<Value> evaluatedResources = new HashSet<>();
+            // Identity-keyed: these accumulate engine values whose hashCode walks the whole
+            // element tree. See HashSetForFhirResourcesAndCqlTypes.
+            final Set<Value> evaluatedResources = new HashSetForFhirResourcesAndCqlTypes<>();
```
lukedegruchy, line 413: Nitpick:  and I didn't notice this on other PRs:  There's a sonar warning on HashSetForFhirResourcesAndCqlTypes:  S2160:  override the equals() method on this class. Might be worth adding an annotation there to ignore that Sonar warning.

antvaset (approved):

lukedegruchy (approved):

PR #1112, opened 2026-09-13
Add tests for evaluating the same library with different versions

antvaset:
- Add equivalent tests for scenarios 1, 2, and 3 from https://github.com/cqframework/clinical_quality_language/pull/1839.
- Add test reproducing the "Could not load source for library LibD, version 1.0.0" exception. This error is expected when LibD version 2.0.0 overwrites LibD version 1.0.0 in the FHIR repository, creating a new version of the same FHIR Library resource.


c-schuler (approved):
Nice work!

One note: consider adding one more negative case for completeness: two distinct-resource-id versions plus a consumer that pins a version not present in the repo, asserting the "could not load" failure

I am not going to block this though

PR #1114, opened 2026-09-16
$create-changelog operation improvments

Chris0296:
$create-changelog output doesn't fully align with the what is expected, when comparing to a manually produced report. Some presentation changes were applied within VSM, but the operation itself needs to be modified to resolve issues of missing data, incorrect attribution and missing change operations.

This catch-all PR resolves problems and shortcomings of the operation, and supports production of a ChangeLog that is fully aligned with expectations.

**Issue**: CodeSystem OID & Name resolution only supported snomed, icd-10, icd-9 and loinc. eRSD contains codes from RXNorm and CVX that are not handled.
**Fix**: Add support for RXNorm and CVX. Create CodeSystemIdentity record as source of truth for CodeSystem resolution.

**Issue**: ChangeLog.addPage used a single code map for both sides, keyed by code value and guarded by !containsKey. A code present in both sides became a single code taken from the source pass, meaning the target side then reported the source side codes version, display and leaf attribution.
**Fix**: Give each side its own code map, allowing codes on each side to carry their own version, display and leaf attribution. 

**Issue**: ChangeLog does not propagate a codes 'inactive' flag field. So it cannot be rendered.
**Fix**: Add Boolean 'inactive' field to ValueSetChild.Code populated from 'expansion.contains.inactive',
reaching compose-claimed codes via expansion fallback.

**Issue**: Conditions and Priority changes not being resolved correctly.
**Fix**: Resolve condition and priority changes by comparing across sides. Diff reports collection membership positionally, so build a per side representation of conditions keyed by leaf, system and code to avoid filling output with positional change data.

**Issue**: Code changes are reported from the grouper level. i.e. If a code is removed from a leaf, but the code is still present in the groupers expansion through another leaf - this is not reported. This does not align with how the manual report is produced, where code changes are reported on a per leaf level.
**Fix**: Store per-side codes keyed by leaf member id, system and version. Compare sides to determine changes rather than using diffs per grouper account of changes.

**Issue**: Code changes where the codes status changed from 'active' to 'inactive', or where the codes description were changed are not reported.
**Fix**: Compare a codes current status/display values to the previous values to determine if changes occurred, add an operation declaring the change if so. 

**Issue**: When a leaf value set carried the same version in both releases, both sides were stored against an identical key and the target resource silently overwrote the source resource. Then when handleValueSetInclude walked the source grouper it would get handed the target's leaf, its expansion, its CodeSystem version and its inactive flags.
**Fix**: Update ArtifactDiffProcessor.DiffCache to have per-side resource maps. 

**Issue**: When a leaf is added or removed, that change is reported - as well as the addition/removal of it's conditions and priorities. This doesn't align with the manual report and generates too much noise.
**Fix**: When a leaf has been added/removed, only report that change. If it hasn't been added/removed, report any changes to it's conditions and priorities.


c-schuler (commented):
More of a general question, but what are the implications for $package/$release? This update is shifting to the use of the `crmi-intendedUsageContext` extension. The `KnowledgeArtifactProcessor` still uses the `vsm-valueset-condition/priority` extensions. So which encoding does the VSM transform actually emit? If `vsm-valueset-condition/priority`, then the changelog finds no conditions/priorities. If `crmi-intendedUsageContext`, then the `KnowledgeArtifactProcessor` is stale. 

Does that make sense?

Chris0296:
> More of a general question, but what are the implications for $package/$release? This update is shifting to the use of the `crmi-intendedUsageContext` extension. The `KnowledgeArtifactProcessor` still uses the `vsm-valueset-condition/priority` extensions. So which encoding does the VSM transform actually emit? If `vsm-valueset-condition/priority`, then the changelog finds no conditions/priorities. If `crmi-intendedUsageContext`, then the `KnowledgeArtifactProcessor` is stale.
> 
> Does that make sense?

@c-schuler 

That makes sense, I hadn't really considered implications for $package/$release.

Looking at VSM, it is emitting with the `crmi-intendedUsageContext` extension shape for conditions and priorities. So it appears the `KnowledgeArtifactProcessor` is stale, I'm not sure how this hasn't come up as an issue. 

I'll need to review to get a better understanding of the drift and implications.

Thanks for raising it!

cqf-fhir-cr/src/main/java/org/opencds/cqf/fhir/cr/crmi/changelog/ValueSetChild.java
c-schuler: Not part of the work you did, but wanted to call out that `.get(urlPart)` is an NPE risk and `next()` does not appear to guarantee "latest"

c-schuler (approved):
Looks good! One comment that is not blocking your work, but would like to at least call it out as a potential issue

sliver007 (approved):

PR #1119, opened 2026-09-19
fix: resolve and extract nested profile answers

rob-reynolds:
Closed as superseded for the requested base conversion fix. The target branch already contains 02873d9069b86aaee8827d2392a7367e0721b682, which handles Coding-to-CodeableConcept conversion with the smaller implementation.

This PR's 2,254 additions and 61 deletions combine that repair with broader profile resolution and nested-container extraction. That broader work is withdrawn from this proposal: focused extraction tests do not establish end-to-end nested-profile support, and failures elsewhere have been reported. It should not be merged to obtain the base conversion fix.

The original implementation remains available in commit 4bb042c7b69af4ba8860d38034450d0b9b3add42. The detailed explanation and test limitations are in https://github.com/cqframework/clinical-reasoning/pull/1119#issuecomment-5744136584. No replacement PR is needed for code already on the target branch.


rob-reynolds:
**Updated scope decision:** The operator reports that nested profiles also fail elsewhere in the system. These extraction changes and their focused tests therefore do not establish usable, end-to-end nested-profile support. Presenting this PR as that fix would be misleading. I withdraw the recommendation to merge the broader nested-profile machinery as part of the base conversion repair. This draft remains unchanged for reference; the smaller upstream conversion fix should stand independently. Any broader proposal needs a concrete end-to-end reproducer and agreement on the affected system layers before implementation is proposed again.

**Exact upstream test result:** Tested unchanged production code at `02873d9069b86aaee8827d2392a7367e0721b682`. Its `ProfileAnswerConversionTests` reports 15 passing cases and two disabled methods. The additional comparison class from this PR reports 15 passing and nine failing cases: the two profile/slice-resolution scenarios and the unsupported-string scenario, each across DSTU3/R4/R5. The string test uses a different target-property spelling from the upstream test, so these are not equivalent assertions. The combined comparison run failed; this is not a claim that the upstream narrow tests failed, nor evidence that this whole PR is required. No full generation/population/extraction round-trip qualification is claimed.

The explanation below describes what the larger patch attempts locally, not a claim that nested profiles work throughout the system.
Scope clarification following the review feedback: I combined the small Coding conversion repair with substantially broader extraction changes. Removing the older branch history in #1119 did not separate those concerns. The broader changes should not be presented as prerequisites for the base conversion fix.

The base branch now contains **02873d9069b86aaee8827d2392a7367e0721b682** (`fix conversion of coding answer to codeable concept profile`). That is the implementation I am evaluating as the base fix; we do not need a competing implementation of that same conversion.

Here is what the extra logic in this PR actually does:

1. **Resolve the precise profile and target element.** The small conversion change operates on an already selected target type. The broader code additionally uses the item's definition canonical and full element/slice identity to select that type. It handles typed aliases, snapshot/differential information, and referenced Extension profiles. The motivating cases are an Extension accidentally borrowing `Observation.value[x]`'s type, or a Quantity slice borrowing a neighboring CodeableConcept constraint. Those are profile-resolution problems, distinct from wrapping a Coding in a CodeableConcept.

2. **Construct complete named paths.** For a definition such as `Observation.extension:outer.extension:inner.value[x]`, the broader code creates the outer and inner containers and attaches the value at the correct depth. It also handles named Observation components. Flat question items and nested questionnaire groups can describe the same target resource tree; this code makes those forms converge on that tree. The conversion fix alone does not construct these ancestors.

3. **Track container identity and repeat scope during one extraction.** The added context/cache reuses a parent when sibling answers belong to the same container, separates repeated group occurrences, preserves enclosing ancestors, and allows an absolute child definition to return to the resource root. This is why the change threads context through several methods. It is broader state coordination, not necessary merely to convert one answer's datatype. Independent answer arrays are not implicitly paired by position.

4. **Apply scoped defaults without corrupting answers.** New containers receive defaults from their own profile/slice. Defaults are copied rather than shared, and revisiting a container does not overwrite previously supplied answers. Extension URLs may come from defaults or answers; URL validation occurs after processing those answers. Answered paths with unresolved required URLs fail visibly, while unanswered paths do not manufacture empty containers. This is additional construction/default behavior.

5. **Retain multiple Coding answers in one CodeableConcept.** The broader assignment logic appends repeated Coding values within the destination CodeableConcept rather than replacing the preceding value. This is a collection-assignment case beyond the single-answer conversion.

6. **Use the root extraction declaration consistently for QR-only requests.** The added root case in `getExtractExtension` uses the same declaration that dispatched definition-based extraction when there is no individual question/response item. That is a separate entry-point correction.

The test files reflect that split: `ProfileAnswerConversionTest` contains the conversion checks plus some boundary cases; `NestedProfileAnswerConversionTest` exercises profile/field identity; `ComplexNamedSliceExtractionTest` exercises resource-tree construction, grouping, repeats and defaults. All are parameterized across DSTU3/R4/R5.

The complete larger patch passed 145 tests with one existing disabled test on e7c5179. That establishes the tested behavior of that larger patch; it does **not** establish that all of it is needed for the production policy. A separate small conversion-only experiment passed 18 of its 24 comparison cases; the six failures were the two profile/slice-resolution cases across the three FHIR versions. The exact upstream results are recorded above; these broader cases are not acceptance requirements for the narrow fix.

The full implementation remains preserved in commit **4bb042c7b69af4ba8860d38034450d0b9b3add42** and the earlier #1108 history. This draft should not be merged wholesale just to obtain the base Coding conversion fix. Any remaining required profile-resolution or nested-construction repair should be discussed and submitted separately with its own concrete reproducer.

PR #1120, opened 2026-09-19
fix: pause ordered applicability on unknown input

rob-reynolds:
An unknown applicability result in an ordered `ANY` group currently allows evaluation to fall through to later actions. This patch pauses at that action, keeps its inputs available for answering, and resumes evaluation when the input is supplied. This is unconditional behavior; there is no setting or opt-out.

Only `ProcessAction.java` and its tests differ from `feature-definition-based-population` at `12f7123`. Applicability is evaluated once per reached action. `false` continues to the next sibling; `true` follows the applicable action; unknown pauses ordered `ANY`. `ALL` siblings remain independent. Evaluation errors and invalid result shapes produce visible OperationOutcome diagnostics, including when another condition is false.

Validation on the final sources:
- Full `cqf-fhir-cr` suite: 1,905 passed, 23 skipped, no failures (includes 17 ProcessAction tests).
- Rebuilt CLI JAR: four native in-memory `$apply` checks passed: initial pause, answer true, change to false, clear and pause again. These check question/answer consistency, no duplicate first question, expected activity, and no error outcomes. This is not JPA/HTTP or customer-policy qualification.
- CLI build and main Checkstyle passed. Changed files are formatting-clean. Repository-wide Spotless fails only on an unused `Disabled` import in unchanged upstream `ProfileAnswerConversionTests.java`; that unrelated file is excluded from this patch.

Draft for review; no nested-profile extraction, generated-ID, duplicate-question, or other upstream fixes are bundled here.


rob-reynolds:
Closing as superseded by the $apply developer's implementation. They have incorporated the useful parts of this proposal and are proceeding with their implementation, so this draft is no longer proposed for merge. Thank you for reviewing it.

